PR #1, opened 2019-09-19
Spring boot starter

benvogler:
This PR now includes a running basic server, with models and services. It also includes a basic react app that is served at `localhost:8080` when compiled or `localhost:3000` when running the react development server.


njd9419:
I'm going to add the M&M Model and Service class to this PR before we merge it.

benvogler:
Sounds good

Looks like travis is failing because it's running java 11. I'll try to figure that out before we merge

PR #5, opened 2019-10-02
Monument Slugs WIP

benvogler:
I've been looking into ways of generating slugs automatically

I work with Salesforce which makes heavy use of database triggers to change data or react to changes on insert/update/delete so that's where my head has immediately gone

Hibernate has these EventListeners such as `PreInsertEventListener` that are basically the same thing, except for some reason way more complex to setup and not-so-great feeling to use.

However, I've got basic slug stuff working with them and there's more left to do:
- [X] Before insert, generate slug
- [X] Before update, modify slug if the related fields (title, city, state) were changed
- [X] Ensure uniqueness by checking if the slug is already used and incrementing a number at the end
- [ ] If there are `slug` and `slug-1`, and `slug` get's deleted/changed, and another matching record is inserted, it becomes `slug-2` even though `slug` is available. Basically this means once a slug is used nothing else can ever take the slug, which is probably not desirable (if there's no `lincoln-memorial` then why should there be a `lincoln-memorial-1` for example)
- [ ] This definitely needs some unit tests
- [ ] Registering listeners is really janky, maybe there's a way to make it nicer
- [ ] Performance concerns - this is doing database queries on every insert/update, how bad does this affect performance when doing bulk inserts/updates

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,7 @@ dependencies {
     runtimeOnly('org.springframework.boot:spring-boot-devtools')
     testImplementation('org.springframework.boot:spring-boot-starter-test')
     implementation('me.alidg:errors-spring-boot-starter:1.2.0')
+    implementation group: 'com.github.slugify', name: 'slugify', version: '2.4'
 }
 
 node {
--- a/src/main/java/com/monumental/listeners/MonumentListener.java
+++ b/src/main/java/com/monumental/listeners/MonumentListener.java
@@ -1,0 +1,20 @@
+package com.monumental.listeners;
+
+import com.monumental.models.Monument;
+import org.hibernate.event.spi.PreInsertEvent;
+import org.hibernate.event.spi.PreInsertEventListener;
+
+public class MonumentListener implements PreInsertEventListener {
+
+    @Override
+    public boolean onPreInsert(PreInsertEvent event) {
+        if (!(event.getEntity() instanceof Monument)) {
+            return false;
+        }
+
+        Monument monument = (Monument) event.getEntity();
+        monument.setSlug("slug");
+
+        return false;
+    }
+}
--- a/src/main/java/com/monumental/listeners/MonumentListener.java
+++ b/src/main/java/com/monumental/listeners/MonumentListener.java
@@ -1,10 +1,17 @@
 package com.monumental.listeners;
 
 import com.monumental.models.Monument;
+import com.monumental.services.MonumentService;
 import org.hibernate.event.spi.PreInsertEvent;
 import org.hibernate.event.spi.PreInsertEventListener;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
+@Component
 public class MonumentListener implements PreInsertEventListener {
+
+    @Autowired
+    MonumentService monumentService;
 
     @Override
     public boolean onPreInsert(PreInsertEvent event) {
@@ -13,7 +20,7 @@ public class MonumentListener implements PreInsertEventListener {
         }
 
         Monument monument = (Monument) event.getEntity();
-        monument.setSlug("slug");
+        monumentService.generateSlug(monument);
 
         return false;
     }
--- a/src/main/java/com/monumental/listeners/MonumentListener.java
+++ b/src/main/java/com/monumental/listeners/MonumentListener.java
@@ -33,22 +33,18 @@ public class MonumentListener implements PreInsertEventListener, PreUpdateEventL
 
     @Override
     public boolean onPreUpdate(PreUpdateEvent event) {
-        try {
-            if (!(event.getEntity() instanceof Monument)) {
-                return false;
-            }
+        if (!(event.getEntity() instanceof Monument)) {
+            return false;
+        }
 
-            Monument updatedMonument = (Monument) event.getEntity();
-            Monument existingMonument = monumentService.get(updatedMonument.getId());
+        Monument updatedMonument = (Monument) event.getEntity();
+        Monument existingMonument = monumentService.get(updatedMonument.getId());
 
-            // This event apparently gets called on insert sometimes, so this skips if there is no existing record
-            if (existingMonument == null) return false;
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
     runtimeOnly('org.springframework.boot:spring-boot-devtools')
     testImplementation('org.springframework.boot:spring-boot-starter-test')
     implementation('me.alidg:errors-spring-boot-starter:1.2.0')
+    implementation group: 'com.github.slugify', name: 'slugify', version: '2.4'
```
benvogler: This is used to do the whole "Lincoln Memorial, Washington D.C." -> "lincoln-memorial-washington-dc" transformation. We could do it ourselves with regex I was just feeling especially lazy today and found this

src/main/java/com/monumental/listeners/MonumentListener.java
```diff
@@ -0,0 +1,52 @@
...
+import com.monumental.models.Monument;
+import com.monumental.services.MonumentService;
+import org.hibernate.event.spi.PreInsertEvent;
+import org.hibernate.event.spi.PreInsertEventListener;
+import org.hibernate.event.spi.PreUpdateEvent;
+import org.hibernate.event.spi.PreUpdateEventListener;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
```
benvogler: `@Component` is used to make Spring find the class and make it a `Bean`, meaning other Spring classes can `@Autowired` it. It's a way to connect this Hibernate stuff to the Spring stuff

src/main/java/com/monumental/listeners/MonumentListener.java
```diff
@@ -0,0 +1,52 @@
...
+    @Autowired
+    MonumentService monumentService;
+
+    /**
+     * Before a Monument is inserted, generate its slug
+     */
+    @Override
+    public boolean onPreInsert(PreInsertEvent event) {
+        // This event is called for all inserts so filter out non-Monuments
+        if (!(event.getEntity() instanceof Monument)) {
```
benvogler: I will try to find a way to make this class only run on insert/update of `Monument`s instead of just checking like this

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -22,5 +32,72 @@ public MonumentService() {
...
+            throw e;
+        }
+
+        // Generate the slug from the Monument's title, city, and state
+        ArrayList<String> slugParts = new ArrayList<>(Arrays.asList(monument.getTitle(), monument.getCity(), monument.getState()));
+        // If these fields do not create a unique key, add on an additional identifier to make it unique
+        if (duplicates.size() > 0) slugParts.add(String.valueOf(duplicates.size()));
+
+        // Use the Slugify package to create a nice, readable slug from these fields
+        monument.setSlug(slugify.slugify(String.join(" ", slugParts)));
```
njd9419: Should we persist a Monument back to the database after we generate its Slug?

benvogler: Actually since you're just changing properties on the object reference, the changes get carried all the way back to hibernate and it completes the transaction with your changes made

benvogler: This is specifically because it's an event before insert and not after

src/main/java/com/monumental/services/SessionFactoryService.java
```diff
@@ -36,4 +42,18 @@ public SessionFactoryService() {
     public SessionFactory getFactory() {
         return this.factory;
     }
+
+    @Autowired
+    private MonumentListener monumentListener;
```
njd9419: Can you move the private variable declaration to the top with `private SessionFactory factory;` ?

src/main/java/com/monumental/listeners/MonumentListener.java
```diff
@@ -0,0 +1,52 @@
...
+            return false;
+        }
+
+        Monument monument = (Monument) event.getEntity();
+        monumentService.generateSlug(monument);
+
+        return false;
```
njd9419: Why return false if we successfully generated the slug?

src/main/java/com/monumental/listeners/MonumentListener.java
```diff
@@ -0,0 +1,52 @@
...
+        Monument existingMonument = monumentService.get(updatedMonument.getId());
+
+        // This event apparently gets called on insert sometimes, so this skips if there is no existing record
+        if (existingMonument == null) return false;
+
+        if (monumentService.slugChanged(existingMonument, updatedMonument)) {
+            monumentService.generateSlug(updatedMonument);
+        }
+
+        return false;
```
njd9419: Same question as above: why return false?

src/main/java/com/monumental/listeners/MonumentListener.java
benvogler: For some reason Hibernate uses this return value to check if the operation should be aborted
https://docs.jboss.org/hibernate/orm/3.5/api/org/hibernate/event/PreInsertEventListener.html

> Return true if the operation should be vetoed

src/main/java/com/monumental/models/Monument.java
```diff
@@ -19,6 +21,9 @@
 })
 public class Monument extends Model implements Serializable {
 
+    @Column(name = "slug", unique = true)
```
njd9419: Since we're utilizing the ID as a way to filter our the existing Monument in some queries, does that mean we're persisting 2 IDs? The auto-incrementing int ID and this ID?

benvogler: Yes in this setup there is both a slug and an id. I read about it a bit and I think this is actually a normal approach because slugs can change and ids can't

benvogler:
After rethinking this a bit I'm going to rewrite this PR so that slugs are only presented to the user and never used on the backend, which means ids will still be in urls but it's not a big deal, i.e.

`localhost:8080/monuments/117/lincoln-memorial-washington-dc`

PR #8, opened 2019-10-04
MonumentService CRUD Unit Tests

njd9419:


build.gradle
```diff
@@ -38,6 +39,8 @@ dependencies {
     runtimeOnly('org.springframework.boot:spring-boot-devtools')
     testImplementation('org.springframework.boot:spring-boot-starter-test')
     implementation('me.alidg:errors-spring-boot-starter:1.2.0')
+    testImplementation('junit:junit:4.12')
+    testImplementation('org.mockito:mockito-core:2.1.0')
```
njd9419: Mockito is the mocking framework I decided to use. Very simple and familiar interface.

src/test/java/com/monumental/services/MonumentServiceTests.java
```diff
@@ -0,0 +1,175 @@
...
+
+        Monument result1 = results.get(0);
+        Monument result2 = results.get(1);
+        Monument result3 = results.get(2);
+
+        assertEquals(this.monument1.getTitle(), result1.getTitle());
+        assertEquals(this.monument2.getTitle(), result2.getTitle());
+        assertEquals(this.monument3.getTitle(), result3.getTitle());
+    }
+}
```
njd9419: I didn't add any unit tests for `MonumentService.update()` or `MonumentService.delete()` because they do not have return values, so there's nothing to assert on. It would basically be testing that the mock does what I tell it to do. These will be covered in the integration tests.

benvogler:
I updated the build.gradle to target java 10 instead of java 12 because apparently what gradle uses to run the tests doesn't support java 12. You do not need java 10 installed for this to work so I think we're all set

src/test/java/com/monumental/services/MonumentServiceTests.java
```diff
@@ -0,0 +1,175 @@
...
+        // Setup SessionFactoryService mock
+        when(this.sessionFactoryService.getFactory()).thenReturn(this.sessionFactory);
+        when(this.sessionFactoryService.getFactory().openSession()).thenReturn(this.session);
+
+        // Setup Session mock
+        when(this.session.beginTransaction()).thenReturn(this.transaction);
+
+        when(this.session.save(this.monument1)).thenReturn(1);
+        when(this.session.save(this.monument2)).thenReturn(2);
+        when(this.session.save(this.monument3)).thenReturn(3);
```
benvogler: I think this is a kind of unfortunate drawback of mockito.

Another way of doing this is with an in-memory database, which lets you actually run SQL and see the results without affecting your local DB https://www.baeldung.com/spring-jpa-test-in-memory-database

PR #13, opened 2019-10-07
React router setup and basic view monument page

benvogler:
This PR:
1. Adds React Router and sets up some basic routing
2. Uses the new header style from our discussion last thursday
3. Adds a basic unformatted view monument page only accessible by URL directly

[Here](https://www.youtube.com/watch?time_continue=122&v=cKnc8gXn80Q)'s a good tutorial on React Router

Here's a screenshot of `localhost:3000/monuments/1`
image

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -1,23 +1,38 @@
 import React from 'react';
 import './App.css';
+import { BrowserRouter as Router, Route } from 'react-router-dom';
+
 import Header from './components/Header/Header';
+import Monument from './components/Monument/Monument';
 
 function App() {
-  return (
-    <div className="App">
-      <Header/>
-      <div className="gmaps">
-        <div className="mapouter">
-          <div className="gmap_canvas">
-            <iframe id="gmap_canvas"
-                    title="gmaps-iframe"
-                    src="https://maps.google.com/maps?q=lincoln%20memorial&t=&z=13&ie=UTF8&iwloc=&output=embed"
-                    frameBorder="0" scrolling="no" marginHeight="0" marginWidth="0"></iframe>
-            Google Maps Generator by <a href="https://www.embedgooglemap.net">embedgooglemap.net</a></div>
+    return (
+        <div className="App">
+            <Router>
+                <Header/>
+                <div style={{padding: '0 4rem'}}>
+                    <Route path="/map">
+                        <div className="gmaps">
+                            <div className="mapouter">
+                                <div className="gmap_canvas">
+                                    <iframe id="gmap_canvas"
+                                            title="gmaps-iframe"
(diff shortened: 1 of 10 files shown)


client/src/App.js
```diff
@@ -1,23 +1,38 @@
...
+                                    Google Maps Generator by <a href="https://www.embedgooglemap.net">embedgooglemap.net</a></div>
+                            </div>
+                        </div>
+                    </Route>
+                    <Route exact path="/">
+                        <div style={{display: 'flex', justifyContent: 'center'}}>
+                            <h1>Welcome!</h1>
+                        </div>
+                    </Route>
+                    <Route path="/monuments/:monumentId" component={Monument}/>
```
benvogler: This is where we route to the monument page

client/src/components/Monument/Monument.js
```diff
@@ -0,0 +1,36 @@
...
+
+    constructor(props) {
+        super(props);
+        this.state = {monument: {}, monumentProperties: []};
+    }
+
+    async componentDidMount() {
+        const { match: { params } } = this.props;
+
+        const response = await fetch(`/api/monument/${params.monumentId}`);
```
benvogler: This is where we access the API

client/src/App.js
njd9419: Maybe I don't understand React enough....why isn't that in a separate file?

client/src/components/Monument/Monument.js
```diff
@@ -0,0 +1,36 @@
...
+import './Monument.css';
+
+export default class Monument extends React.Component {
+
+    constructor(props) {
+        super(props);
+        this.state = {monument: {}, monumentProperties: []};
+    }
+
+    async componentDidMount() {
```
njd9419: Why is this method called `componentDidMount()`? Is this a standard or something?

client/src/App.js
benvogler: What do you mean? It is in a separate file, it's rendering the `Monument` component which is `components/Monument/Monument.js`

client/src/components/Monument/Monument.js
benvogler: Yes, it's part of the component lifecycle. This basically is overriding `React.Component`'s method, since this class extends that one

client/src/App.js
njd9419: So our App.js will be updated with new Routes each time we add one?

benvogler: We could keep it in App.js or make a specific component for it, but I think it makes sense to do it right in App.js (in the future, the routes will probably all render a specific component, instead of some html direct in App.js like I currently have for the interactive map, so it'll be more compact and clearer)

PR #16, opened 2019-10-07
Add 404 Error for /api/monuments/{id}

### LINKED ISSUE #9: Add 404 errors to the API
Currently if you do a GET to `/api/monument/<id that doesn't exist>` you just get an empty JSON response. This should be responding with a 404 error instead

```java
import me.alidg.errors.annotation.ExceptionMapping;
import org.springframework.http.HttpStatus;

@ExceptionMapping(statusCode = HttpStatus.NOT_FOUND, errorCode = "notFound")
public class NotFoundException extends RuntimeException
```

njd9419:
This PR adds custom error handling and messages to the /api/monuments/{id} endpoint.

When a request attempts to retrieve a Monument that does not exist, the response will now be a 404 with a detailed error message.

This closes #9.


src/main/resources/messages/monuments.properties
```diff
@@ -0,0 +1 @@
+monument.not_found=No Monument with the specified id found
```
benvogler: If we're not referencing anything specific to monuments here maybe we can change `MonumentNotFoundException` to be generic and reuse it everywhere 🤷‍♂

njd9419: Yeah, might as well. I thought maybe having the message for each record kind different but 1. What's the point and 2. Who knows how many other records we'll have anyway.

PR #17, opened 2019-10-07
Fix issues in CsvMonumentConverter

njd9419:
This PR fixes 2 issues in `CsvMonumentConverter.java`:

1. Fixes the incorrect latitude and longitude values (they were swapped)
2. Removes quotes from fields

### DIFF
--- a/src/main/java/util/csvparsing/CsvMonumentConverter.java
+++ b/src/main/java/util/csvparsing/CsvMonumentConverter.java
@@ -21,8 +21,8 @@ public class CsvMonumentConverter {
         Monument result = new Monument();
 
         for (int columnIndex = 0; columnIndex < csvRowArray.length; columnIndex++) {
-            // Grab the value at the current column and remove any quotes that may be left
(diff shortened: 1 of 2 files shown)


src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -21,7 +21,8 @@ public static Monument convertCsvRowToMonument(String csvRow) {
         Monument result = new Monument();
 
         for (int columnIndex = 0; columnIndex < csvRowArray.length; columnIndex++) {
-            String value = csvRowArray[columnIndex];
+            // Grab the value at the current column and remove any quotes that may be left
+            String value = csvRowArray[columnIndex].replace("\"", "");
```
benvogler: what if the title is actually supposed to have quotes in it? this should probably only check for quotes at the beginning & end

njd9419: Yup, working on that right now

PR #19, opened 2019-10-09
Framework for database triggers, groundwork for geocoding addresses

benvogler:
This PR does not include any integration with the Google Maps API

The main purpose of this PR is to establish the `ModelTrigger` class, which is a generic parent class similar to `Model` and `ModelService` (although this one is not abstract like `ModelService`, it still should not be used without extending it first)

The class essentially listens to the following events:
1. Pre-Insert
2. Post-Insert
3. Pre-Update
4. Post-Update
5. Pre-Delete
6. Post-Delete

There are other Hibernate events that it could listen to but none seemed relevant for now

Once a class extends the `ModelService` it will have an associated generic type for the `Model` it corresponds with (`Monument` in `MonumentTrigger`'s case).

Then it filters the events it listens to, only listening to the ones for that `Model`. When it gets a hit it calls the corresponding method such as `beforeUpdate`, `afterDelete` etc. These methods are skeletons until the subclass overrides them with some functionality. If the subclass does not override them then nothing happens and the transaction continues on as normal.

If the subclass chooses to override the method, say `beforeInsert`, then that method is called and runs before the transaction continues. In `beforeInsert`, changes made to the `record` parameter are kept and affect the upcoming SQL insert, while in a `afterInsert` the changes made have no effect and you'd have to call an update to persist them.

For `beforeUpdate` and `afterUpdate` the intention is to have the same behavior but there seems to be a bug in Hibernate where changes made in `beforeUpdate` do not persist. I found a thread on it [here](https://forum.hibernate.org/viewtopic.php?f=1&t=1042825&view=next) and a workaround [here](http://anshuiitk.blogspot.com/2010/11/hibernate-pre-database-opertaion-event.html), the solution basically being that you have to call the method `setProperty("propertyName", someValue)` and then that change will persist when the update goes through.

### DIFF
--- a/src/main/java/com/monumental/controllers/MonumentController.java
+++ b/src/main/java/com/monumental/controllers/MonumentController.java
@@ -1,10 +1,12 @@
-package com.monumental.controllers.api;
+package com.monumental.controllers;
 
 import com.monumental.services.exceptions.ResourceNotFoundException;
 import com.monumental.models.Monument;
 import com.monumental.services.MonumentService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
 
 @RestController
 public class MonumentController {
@@ -26,6 +28,17 @@ public class MonumentController {
             throw new ResourceNotFoundException();
         }
 
+        return monument;
+    }
+
+    @GetMapping("/api/monuments")
+    public List<Monument> getAllMonuments() {
+        return this.monumentService.getAll();
+    }
+
+    @PutMapping("/api/monument/{id}")
+    public Monument updateMonument(@PathVariable("id") Integer id, Monument monument) {
+        this.monumentService.update(monument);
         return monument;
     }
 }
--- a/src/main/java/com/monumental/models/Monument.java
+++ b/src/main/java/com/monumental/models/Monument.java
@@ -36,16 +36,19 @@ public class Monument extends Model implements Serializable {
     private String material;
 
     @Column(name = "lat")
-    private double lat;
+    private Double lat;
 
     @Column(name = "lon")
-    private double lon;
+    private Double lon;
 
     @Column(name = "city")
     private String city;
 
     @Column(name = "state")
     private String state;
+
+    @Column(name = "address")
+    private String address;
 
     public Monument() {
 
@@ -121,6 +124,7 @@ public class Monument extends Model implements Serializable {
     }
 
     public String getCoordinatePointAsString() {
+        if (this.lat == null || this.lon == null ) return null;
         return Double.toString(this.lat) + ", " + Double.toString(this.lon);
     }
 
@@ -140,9 +144,17 @@ public class Monument extends Model implements Serializable {
         this.state = state;
     }
 
+    public String getAddress() {
+        return this.address;
+    }
+
+    public void setAddress(String address) {
+        this.address = address;
+    }
+
     public String toString() {
         return "Submitted by: " + this.submittedBy +  ", Artist: " + this.artist + ", Title: " + this.title + ", Date: "
                 + this.date + ", Material: " + this.material + ", Coordinates: " + this.getCoordinatePointAsString()
-                + ", City: " + this.city + ", State: " + this.state;
+                + ", City: " + this.city + ", State: " + this.state + ", Address: " + this.address;
     }
 }
--- a/src/main/java/com/monumental/services/ModelService.java
+++ b/src/main/java/com/monumental/services/ModelService.java
@@ -77,7 +77,7 @@ public abstract class ModelService<T extends Model> {
             } else {
                 records = session.createQuery(
                     "FROM " + tableName +
-                    "WHERE id IN (:ids)"
+                    " WHERE id IN (:ids)"
                 ).setParameter("ids", ids).list();
             }
             transaction.commit();
--- a/src/main/java/com/monumental/services/ModelService.java
+++ b/src/main/java/com/monumental/services/ModelService.java
@@ -8,9 +8,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.lang.reflect.ParameterizedType;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
+import java.util.*;
 
 /**
  * This class handles the generic CRUD of Hibernate entities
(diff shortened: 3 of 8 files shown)


src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -21,4 +23,15 @@ public Monument createMonument(Monument monument) {
...
         return this.monumentService.get(id);
     }
+
+    @GetMapping("/api/monuments")
+    public List<Monument> getAllMonuments() {
+        return this.monumentService.getAll();
+    }
+
+    @PutMapping("/api/monument/{id}")
+    public Monument updateMonument(@PathVariable("id") Integer id, Monument monument) {
```
benvogler: This method completely replaces the row in the database with whatever JSON you provide. If you don't include a field it will be set to null even if some data exists for it in the database. I think in all cases React will have all the data so this isn't an issue, but keep in mind while testing

src/main/java/com/monumental/models/Monument.java
```diff
@@ -36,17 +36,20 @@
     private String material;
 
     @Column(name = "lat")
-    private double lat;
+    private Double lat;
 
     @Column(name = "lon")
-    private double lon;
+    private Double lon;
```
benvogler: This change is to make the fields nullable, since `double` defaults to `0.0` instead of `null`

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -66,16 +64,21 @@
...
+                records = new ArrayList<>();
+                records.add((T) session.get(this.getModelClass(), ids.get(0)));
             } else {
-                for (Integer id : ids) {
-                    records.add((T) session.get(this.getModelClass(), id));
-                }
+                records = session.createQuery(
+                    "FROM " + tableName +
+                    "WHERE id IN (:ids)"
+                ).setParameter("ids", ids).list();
```
benvogler: This theoretically makes querying for lists of ids more efficient since it's 1 query instead of N

src/main/java/com/monumental/services/SessionFactoryService.java
```diff
@@ -1,28 +1,52 @@
...
 
+import javax.annotation.PostConstruct;
+import java.util.Arrays;
+import java.util.List;
+
 @Service
 public class SessionFactoryService {
 
+    @Autowired
+    public MonumentTrigger monumentTrigger;
```
benvogler: These need to be autowired so that they can call other services if need be

benvogler: They also have to be public due to some reflection nonsense farther down

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -21,4 +23,15 @@ public Monument createMonument(Monument monument) {
     public Monument getMonument(@PathVariable("id") Integer id) {
         return this.monumentService.get(id);
     }
+
+    @GetMapping("/api/monuments")
+    public List<Monument> getAllMonuments() {
+        return this.monumentService.getAll();
+    }
+
+    @PutMapping("/api/monument/{id}")
```
njd9419: `@PutMapping` instead of `@PostMapping`?

Are we using PUT for updates and POST for creation?

src/main/java/com/monumental/services/SessionFactoryService.java
```diff
@@ -1,28 +1,52 @@
...
+    /** IMPORTANT: Update these lists when creating triggers or models
+     *  Triggers must be autowired, and the string must match the property's name
+     */
+    private final List<String> triggers = Arrays.asList(
+        "monumentTrigger"
+    );
+    private final List<Class> models = Arrays.asList(
+        Monument.class
+    );
+
```
njd9419: 00f.

src/main/java/com/monumental/services/SessionFactoryService.java
```diff
@@ -1,28 +1,52 @@
...
+        EventType.PRE_INSERT, EventType.POST_INSERT, EventType.PRE_UPDATE,
+        EventType.POST_UPDATE, EventType.PRE_DELETE, EventType.POST_DELETE
+    };
+
     public SessionFactoryService() {
 
         try {
             Configuration conf = new Configuration()
                     .configure();
-            Class[] classes = new Class[]{
```
njd9419: Since we removed this Array, aren't we now missing `Model.class`? It's not added to `this.models`

src/main/java/com/monumental/services/SessionFactoryService.java
```diff
@@ -36,4 +60,23 @@ public SessionFactoryService() {
...
+     * Listeners must be registered here so that they are called for the proper events
+     * If a listener listens to multiple EventTypes it must be added for each EventType
+     */
+    @SuppressWarnings("unchecked")
+    @PostConstruct
+    public void registerListeners() throws NoSuchFieldException, IllegalAccessException {
+        EventListenerRegistry registry = ((SessionFactoryImpl) this.factory).getServiceRegistry().getService(EventListenerRegistry.class);
+        // Make every ModelTrigger execute on every EventType
+        // If the ModelTrigger doesn't override the associated methods then they won't do anything
+        for (EventType type : eventTypes) {
```
njd9419: `eventTypes` --> `this.eventTypes` :)

src/main/java/com/monumental/triggers/ModelTrigger.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /** Hibernate Event Listeners */
+
+    @Override
+    public boolean onPreInsert(PreInsertEvent event) {
+        Object entity = event.getEntity();
+        if (entity.getClass().equals(this.getModelClass())) {
+            this.beforeInsert((T) entity);
+        }
+        return false;
```
njd9419: Why return `false`?

src/main/java/com/monumental/triggers/ModelTrigger.java
```diff
@@ -0,0 +1,135 @@
...
+ * This class wraps the Hibernate event system into an easier-to-use parent class
+ * @param <T> The Model to trigger on - only this Model type will be watched by the subclass
+ */
+@SuppressWarnings("unchecked")
+@Component
+public class ModelTrigger<T extends Model> implements
+        PreInsertEventListener, PostInsertEventListener, PreUpdateEventListener, PostUpdateEventListener,
+        PreDeleteEventListener, PostDeleteEventListener {
+
+    private List<String> propertyNames;
```
njd9419: Any reason to just have these as local variables?

src/test/java/com/monumental/services/MonumentServiceTests.java
```diff
@@ -155,11 +156,11 @@ public void unitTestMonumentServiceGetAll_ListPassed() {
...
 
     /**
-     * Test method for unit testing MonumentService.getAll(null)
+     * Test method for unit testing MonumentService.getAll()
      * Mocks the appropriate classes as to not connect to the database
      */
     @Test
-    public void unitTestMonumentServiceGetAll_NullPassed() {
+    public void unitTestMonumentServiceGetAll() {
         List<Monument> results = this.monumentService.getAll();
```
njd9419: Wasn't `MonumentService.getAll()` removed and replaced with `MonumentService.get(null)`?

src/main/java/com/monumental/services/SessionFactoryService.java
benvogler: I think we might be able to avoid this in the future using spring's annotation scanning https://stackoverflow.com/a/1415338/10044594

src/main/java/com/monumental/services/SessionFactoryService.java
benvogler: good catch

src/main/java/com/monumental/triggers/ModelTrigger.java
benvogler: Returning `true` vetoes the operation (prevents the insert) https://docs.jboss.org/hibernate/orm/4.1/javadocs/org/hibernate/event/spi/PreInsertEventListener.html

I'll add a comment about this

benvogler:
@njd9419 the tests are failing for this because I changed the `get` query to use a list of ids and now I can't get the mock for it to work right

benvogler:
@njd9419 see my comment above yours lol

njd9419:
Lol ok I'll take a look at it.

PR #22, opened 2019-10-10
Add embedded coordinate map to view monument pages

benvogler:
I'm not really sure how, but this way of embedding the map avoids using an API key entirely.

image

### DIFF
--- a/client/src/components/Monument/Monument.js
+++ b/client/src/components/Monument/Monument.js
@@ -13,6 +13,7 @@ export default class Monument extends React.Component {
 
         const response = await fetch(`/api/monument/${params.monumentId}`);
         const monument = await response.json();
+        console.log(monument);
         const monumentProperties = [];
         for (let prop in monument) {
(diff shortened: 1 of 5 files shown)


client/src/components/Monument/Monument.js
```diff
@@ -9,27 +11,53 @@ export default class Monument extends React.Component {
     }
 
     async componentDidMount() {
-        const { match: { params } } = this.props;
-
-        const response = await fetch(`/api/monument/${params.monumentId}`);
+        const { match: { params: { monumentId, slug } } } = this.props;
+        const response = await fetch(`/api/monument/${monumentId}`);
         const monument = await response.json();
+        console.log(monument);
```
njd9419: Debug comment?

benvogler: Intentionally left it in, it will be nice to have easy access to that info going forward. I could add a todo to get rid of it eventually

PR #23, opened 2019-10-11
Monument -- Add groundwork for Image and Tag support

njd9419:
I added a Tag table and Image table that we can use in future work.

Tags and Monuments have a many-to-many relationship. At the database level, this is done via a join table with two columns: monument_id and tag_id. Both columns are foreign keys to the monument table and tag table respectively.

Images and Monuments have a many-to-one relationship. At the database level, this is just a join column in the images table that is a foreign key to the id column in the monument table.

### DIFF
--- a/src/main/java/com/monumental/models/Image.java
+++ b/src/main/java/com/monumental/models/Image.java
@@ -1,0 +1,44 @@
+package com.monumental.models;
+
+import javax.persistence.*;
+import java.io.Serializable;
+
+/**
+ * Model class for an Image
+ * Contains all of the state related to an Image and Getters and Setters for that state
+ * Contains a many-to-one relationship with Monument
+ */
+
+@Entity
+@Table(name = "image", uniqueConstraints = {
+        @UniqueConstraint(columnNames = "id")
+})
+public class Image extends Model implements Serializable {
+
+    @Column(name = "url")
+    private String url;
+
+    @ManyToOne
+    @JoinColumn(name = "monument_id", nullable = false)
+    private Monument monument;
+
+    public Image() {
+
+    }
+
+    public String getUrl() {
+        return this.url;
+    }
+
+    public void setUrl(String url) {
+        this.url = url;
+    }
+
+    public Monument getMonument() {
+        return this.monument;
+    }
+
+    public void setMonument(Monument monument) {
+        this.monument = monument;
+    }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/com/monumental/models/Image.java
```diff
@@ -0,0 +1,44 @@
...
+@Entity
+@Table(name = "image", uniqueConstraints = {
+        @UniqueConstraint(columnNames = "id")
+})
+public class Image extends Model implements Serializable {
+
+    @Column(name = "url")
+    private String url;
+
+    @ManyToOne
```
benvogler: We should use lazy loading on these since we won't always need them
```suggestion
    @ManyToOne(fetch = FetchType.LAZY)
```

src/main/java/com/monumental/models/Monument.java
```diff
@@ -46,10 +45,21 @@
 
     @Column(name = "state")
     private String state;
-
+  
     @Column(name = "address")
     private String address;
 
+    @ManyToMany(cascade = { CascadeType.ALL })
```
benvogler:
```suggestion
    @ManyToMany(cascade = { CascadeType.ALL }, fetch = FetchType.LAZY)
```

src/main/java/com/monumental/models/Monument.java
```diff
@@ -46,10 +45,21 @@
...
+    @JoinTable(
+            name = "monument_tag",
+            joinColumns = { @JoinColumn(name = "monument_id", referencedColumnName = "id") },
+            inverseJoinColumns = { @JoinColumn(name = "tag_id", referencedColumnName = "id") }
+    )
+    private Set<Tag> tags;
+
+    @OneToMany(mappedBy = "monument")
```
benvogler:
```suggestion
    @OneToMany(mappedBy = "monument", fetch = FetchType.LAZY)
```

src/main/java/com/monumental/models/Tag.java
```diff
@@ -0,0 +1,44 @@
...
+@Entity
+@Table(name = "tag", uniqueConstraints = {
+        @UniqueConstraint(columnNames = "id")
+})
+public class Tag extends Model implements Serializable {
+
+    @Column(name = "name")
+    private String name;
+
+    @ManyToMany(mappedBy = "tags")
```
benvogler:
```suggestion
    @ManyToMany(mappedBy = "tags", fetch = FetchType.LAZY)
```

benvogler:
https://www.baeldung.com/hibernate-lazy-eager-loading

PR #26, opened 2019-10-12
Add address lookup via reverse geocoding

### LINKED ISSUE #21: POST /api/monument uses query params instead of JSON body
Pretty simple mistake, we just need to change:
```java
@PostMapping("/api/monument")
public Monument createMonument(Monument monument) {
```
to
```java
@PostMapping("/api/monument")
public Monument createMonument(@RequestBody Monument monument) {
```

benvogler:
This PR:
1. Adds automatic address lookup in google maps via reverse geocoding when a monument has no address or the lat/lon changes.
2. Closes #21

I still need to run this on all the records and get the data to everyone so we don't waste our budget sending those requests multiple times.

Also, the API key is currently just pulled from environment variables which you can set by editing your run configuration in IntelliJ. For now I'm just going to hold onto the API key


src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -37,8 +37,14 @@ public Monument getMonument(@PathVariable("id") Integer id) throws ResourceNotFo
...
     @PutMapping("/api/monument/{id}")
-    public Monument updateMonument(@PathVariable("id") Integer id, Monument monument) {
-        this.monumentService.update(monument);
+    public Monument updateMonument(@PathVariable("id") Integer id, @RequestBody Monument monument) {
+        monument.setId(id);
+        System.out.println("in put " + monument.toString());
+        try {
+            this.monumentService.update(monument);
+        } catch (Exception e) {
+            e.printStackTrace();
```
njd9419: Debug print still left?

benvogler: Intentional

benvogler: oh wait line 42. never mind I'll remove that

PR #27, opened 2019-10-12
Added New Classes and Attributes for View M&M

njd9419:
This PR adds the new classes and attributes as discussed in the 10/10 meeting.

It adds:

1. A Contribution Model class with a one-to-many relationship with Monument
2. A Reference Model class with a one-to-many relationship with Monument
3. A Description attribute to the Model class
4. An Inscription attribute to the Model class

I also removed the submittedBy attribute from Monument as that will instead be added as a Contribution on the Monument.

We will store null for the Description attribute in the database. When getDescription() is called, it checks if the attribute is null or not. If it is, it generates a description and returns that. This is not stored in the database. If the attribute is not null, it just returns whatever was in the database.


src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -29,7 +29,7 @@ public static Monument convertCsvRowToMonument(String csvRow) {
             // I also assumed that if any columns were empty, they would be null
             switch (columnIndex) {
                 case 0: // Submitted By
-                    result.setSubmittedBy(value);
+                    //result.setSubmittedBy(value);
```
njd9419: Planning to fix this class under the other task that is currently assigned to me.

njd9419:
I'm going to write some quick tests for this branch before it gets merged. Please don't merge it until after I'm done.

src/main/java/com/monumental/models/Contribution.java
```diff
@@ -0,0 +1,56 @@
+package com.monumental.models;
+
+import javax.persistence.*;
+import java.io.Serializable;
+import java.util.Date;
+
+/**
+ * Model class for a Contribution
+ * Contains all of the state associated with a Contribution along with Getters and Setters for that state
+ */
```
benvogler: Random thought: she may want:
1. Users to have a public profile to show their contributions
2. Users to have a private page with the same info

Not saying we should promise any of that now, but it makes sense down the road once users exist

src/main/java/com/monumental/models/Monument.java
```diff
@@ -49,24 +46,35 @@
...
 
     public Monument() {
 
     }
 
-    public Monument(String submittedBy, String artist, String title, Date date, String material, Double lat,
-                    Double lon, String city, String state) {
-        this.submittedBy = submittedBy;
+    public Monument(String artist, String title, Date date, String material, double lat,
+                    double lon, String city, String state) {
```
benvogler: One of the things I hate about java is the lack of dynamic constructors. In Salesforce's shitty version of Java you can do
```java
new Monument(artist="Bob", title="Some Monument");
```
and it's super convenient. To my knowledge there's no way to achieve that in Java other than accepting a list/set as a param which sucks because then it isn't typed/doesn't check that the fields exist

benvogler: This rant was brought to you by Sleep Deprevation®

src/main/java/com/monumental/models/Monument.java
```diff
@@ -162,25 +162,88 @@ public void setAddress(String address) {
...
+     * @return String - the description of this Monument
+     */
+    private String generateDescription() {
+        String description = "";
+
+        if (!this.title.toLowerCase().startsWith("the ")) {
+            description += "The ";
+        }
+
+        description += this.title + " in " + this.city + ", " + this.state + " was created by " + this.artist + " in ";
```
benvogler: Does she want material in here?

src/main/java/com/monumental/services/SessionFactoryService.java
```diff
@@ -33,7 +30,7 @@
     );
     private final List<Class> models = Arrays.asList(
         Model.class, Monument.class, Tag.class,
-        Image.class
+        Image.class, Contribution.class, Reference.class
```
benvogler: I really need to make this reflective, bobby would not be happy

src/test/java/com/monumental/models/unittest/MonumentUnitTests.java
```diff
@@ -0,0 +1,165 @@
...
+        Monument monument = makeTestMonument("The Title", "City", "State", "Artist", new Date(), null);
+
+        String result = monument.getDescription();
+
+        assertEquals("The Title in City, State was created by Artist in 2019.", result);
+    }
+
+    @Test
+    public void testMonument_getDescription_TitleStartsWithThe_MixedCase() {
+        Monument monument = makeTestMonument("ThE Title", "City", "State", "Artist", new Date(), null);
```
benvogler: image

benvogler:
💯

src/main/java/com/monumental/models/Monument.java
njd9419: That's super nice.

Java also doesn't have freaking default parameters which is aids.

PR #28, opened 2019-10-13
Add MonumentService CRUD Integration Tests

njd9419:
This PR adds integration tests for the MonumentService CRUD methods. Setting up H2 took very little time so I decided to just go ahead and do it.


src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
```diff
@@ -0,0 +1,298 @@
...
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+@SpringBootTest
+@RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
njd9419: This Spring annotation makes sure that the database is reset before each test method runs so that we can know what is in there and what IDs to expect on inserts.

src/test/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version = "1.0" encoding = "utf-8"?>
```
njd9419: Adding H2 was as simple as adding a separate hibernate.cfg.xml file in src/test/resources that uses the H2 dialect and driver class.

njd9419:
The integration tests are currently failing because they are trying to do the pre-insert and pre-update MonumentTrigger actions even during tests.

Trying to find a way to ignore these during testing.

src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
```diff
@@ -0,0 +1,298 @@
...
+                assertEquals(monuments.get(2).getArtist(), result.getArtist());
+            }
+        }
+    }
+
+    /**
+     * Test method for integration testing MonumentService.getAll(null)
+     * First inserts 3 records into the database then performs a getAll for all 3 records
+     * Checks that 3 records are returned and that they have the expected attributes
+     * Uses an H2 in-memory database as to not ruin the actual database
```
benvogler: I don't think we need this comment on all of them. Putting it on the class would suffice

src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
```diff
@@ -0,0 +1,298 @@
...
+        ArrayList<Monument> monuments = makeTestMonuments();
+
+        this.monumentService.insert(monuments);
+
+        List<Monument> results = this.monumentService.getAll();
+
+        assertEquals(3, results.size());
+
+        for (Monument result : results) {
+            if (result.getId() == 1) {
```
benvogler: I'm being pedantic here, but you could create a map of Id to monument and avoid any if statements here. I do think that you're right that the order in which `getAll` currently makes no sense and should be ordered by _something_, probably Id for now

benvogler: That being said even if it returns ordered by Id I don't think our tests should rely on the order in which results are returned

benvogler:
@njd9419 I just pushed some changes that makes the `GOOGLE_API_KEY` optional, if it's not specified no calls are made to google and methods will just return null or similar.

benvogler:
@njd9419 getting this error: `org.hibernate.PessimisticLockException: could not extract ResultSet`

I think this is an issue related to the join table you created for tags: https://stackoverflow.com/questions/20089031/could-not-extract-resultset-in-hibernate

src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
njd9419: Fixed.

src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
njd9419: Agreed that at some point we should definitely order `getAll()` by id.

njd9419: Will be fixed in the next commit.

njd9419:
Looking into the tests failing now...

src/test/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,23 @@
...
+        <property name="hibernate.dialect">
+            org.hibernate.dialect.H2Dialect
+        </property>
+
+        <property name="hibernate.connection.driver_class">
+            org.h2.Driver
+        </property>
+
+        <property name="hibernate.connection.url">
+            jdbc:h2:mem:postgres;DB_CLOSE_DELAY=-1;MVCC=true
```
njd9419: Adding `MVCC=true` fixed the failing tests locally for me.

This is enabling Multi-Version Concurrency Control for H2.

Find details on it here: https://stackoverflow.com/questions/4162557/timeout-error-trying-to-lock-table-in-h2 and here: http://www.h2database.com/html/advanced.html#mvcc

benvogler: @njd9419 I'm getting this error outside of the test environment now when I query for contributions/images, and this fix only works for h2

benvogler: Actually it's the references that's doing it:
>  Resolved [org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: com.monumental.models.Monument.references, could not initialize proxy - no Session; nested exception is com.fasterxml.jackson.databind.JsonMappingException: failed to lazily initialize a collection of role: com.monumental.models.Monument.references, could not initialize proxy - no Session (through reference chain: com.monumental.models.Monument["description"])]

benvogler: Never mind, I just need to add @JSONIgnore again

PR #32, opened 2019-10-18
Update CsvMonumentConverter and InitialDataLoadApplication for New Fields

njd9419:
Finally getting this into PR. There were a few things that came up that prevented me from getting it up sooner.

This PR:
- Parses the additional columns in the Initial Dataset Spreadsheet to fill in the new attributes we discussed: References, Contributions, Inscriptions and Tags
- Adds the groundwork for Monument (and the broader Model) validation and exposes it in a manual manner if needed
- Adds a bunch of unit and integration tests around the new things implemented

When you go to run this on to load your local database, be sure to download and use the Initial Dataset.csv file that is in the Drive, inside the Data Sets from Dr. Decker folder. This CSV file is scrubbed into a almost 100% acceptable state so that the data we have is valid.


src/main/java/com/monumental/models/Contribution.java
```diff
@@ -18,6 +19,7 @@
 public class Contribution extends Model implements Serializable {
 
     @Column(name = "submitted_by")
+    @NotNull(groups = NewOrExisting.class, message = "Submitted By can not be null")
```
njd9419: `@NotNull` is one of the JPA constraint annotations and allows for a custom message when violated.

src/main/java/com/monumental/models/Model.java
```diff
@@ -4,14 +4,38 @@
...
+    // Use this group to signify validation that occurs on record creation or after record creation
+    public interface NewOrExisting extends New, Existing {
+
+    }
+
+    // Use this group to signify validation of the Id field of the record
+    public interface Identity {
+
+    }
+
```
njd9419: These are groups we can use on our `@NotNull` and other JPA constraint annotations to signify when we want the validation to occur.

src/main/java/com/monumental/models/Monument.java
```diff
@@ -258,4 +267,82 @@ private String generateDescription() {
...
+        this.references.add(reference);
+    }
+
+    /**
+     * Encapsulates the logic to validate a Monument object
+     * Use this method to manually run validation in lieu of a @Valid Spring annotation
+     * @return List<String> - List of ConstraintViolation messages, if any
+     * @return null - If there are no ConstraintViolations
+     */
+    public List<String> validate() {
```
njd9419: This is a method that we can use to manually validate a Monument object instead of relying on an `@Valid` annotation to do it for us.

Since Java doesn't have `out`, the return of this method isn't what I'd want it to be (returns a boolean and outs a List of exception messages) so it's on the caller to get if what is returned is null or not to determine if the Monument is valid.

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -281,6 +281,39 @@ public void delete(List<Integer> ids) throws HibernateException {
         return records;
     }
```
njd9419: This is a generic method we can use in the future to make queries with `WHERE` clauses on our tables without having to repeat all of the other junk.

src/main/java/util/InitialDataLoadApplication.java
```diff
@@ -35,23 +41,61 @@ public static void main(String[] args) {
...
+                if (r.getTags() != null) {
+                    for (Tag t : r.getTags()) {
+                        try {
+                            tagService.insert(t);
+                            tagInsertedCount++;
+                        } catch (ConstraintViolationException e) {
+                            // If a ConstraintViolationException is caught, there's a duplicate Tag in the list we are trying
+                            // to insert
+                            // Instead, get the ID of the already existing Tag and update it
+                            // Assume there is only 1 Tag with that Name due to the Unique Constraint
```
njd9419: Having to catch a `ConstraintViolationException`, grabbing the duplicate's ID, setting it then updating isn't ideal. Perhaps in the future we can change `insert()` to encapsulate this logic so the caller doesn't have to worry about it.

src/main/java/com/monumental/models/Monument.java
benvogler: Instead of using `out` you could just make a class (within this class) as a data holder and return that

njd9419: Now makes use of `Monument.ValidationResult`

src/main/java/com/monumental/models/Model.java
```diff
@@ -4,14 +4,38 @@
...
+    // Use this group to signify validation that occurs on record creation or after record creation
+    public interface NewOrExisting extends New, Existing {
+
+    }
+
+    // Use this group to signify validation of the Id field of the record
+    public interface Identity {
```
benvogler: I know I probably wrote this, but I really don't understand what this is for

src/main/java/com/monumental/models/Monument.java
```diff
@@ -52,23 +60,23 @@
     @Column(name = "description")
     private String description;
 
-    @Column(name = "inscription")
+    @Column(name = "inscription", length = 1024)
```
benvogler: what does this do? make the max length 1024 chars? if so that's probably too short

src/main/java/com/monumental/models/Monument.java
```diff
@@ -249,7 +257,8 @@ private String generateDescription() {
                 Reference firstReference = this.references.get(0);
 
                 if (firstReference != null && firstReference.getUrl() != null) {
-                    description += " You may find further information about this monument at: " + firstReference.getUrl();
+                    description += " You may find further information about this monument or memorial at: " +
+                            firstReference.getUrl();
```
benvogler: Sort of related to the image gallery we may want a way to signify a primary reference rather than just take the oldest

src/main/java/com/monumental/models/Monument.java
```diff
@@ -258,4 +267,108 @@ private String generateDescription() {
...
+
+    /**
+     * Adds a Contribution to the List
+     * Will make a new ArrayList if this.contributions is null
+     * Checks if a Contribution has already been added to this.contributions with the same date and name
+     * Will do nothing if the specified Contribution has a null date or null/empty submittedBy
+     * and does nothing if so
+     * @param contribution - Contribution to add to the List
+     */
+    public void addContribution(Contribution contribution) {
```
benvogler: These are covered by the `NotNull` and such not aren't they? do we still need these? I think this is bad design if we do need them

src/main/java/com/monumental/models/Monument.java
```diff
@@ -258,4 +267,108 @@ private String generateDescription() {
...
+
+        this.references.add(reference);
+    }
+
+    /**
+     * Encapsulates the logic to validate a Monument object
+     * Use this method to manually run validation in lieu of a @Valid Spring annotation
+     * @return ValidationResult - ValidationResult object representing the result of the validation
+     */
+    public ValidationResult validate() {
```
benvogler: Why is this needed?

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -281,6 +281,39 @@ public void delete(List<Integer> ids) throws HibernateException {
...
 
+    /**
+     * Generic method to build a "FROM ... WHERE ..." query
+     * Assumes the FROM clause is the name of the T generic class
+     * @param whereParameterName - Name of the column to use in the WHERE clause
+     * @param whereParameter - Used to set the parameter in the WHERE clause
+     * @return List<T> - the List of T returned by the query
+     */
+    @SuppressWarnings("unchecked")
+    public List<T> getFromWhere(String whereParameterName, Object whereParameter) {
```
benvogler: This should also have an option to initialize all the lazy loaded lists like in the other gets

src/main/java/com/monumental/services/TagService.java
```diff
@@ -8,11 +8,36 @@
 @Service
 public class TagService extends ModelService<Tag> {
 
+    /**
+     * Public constructor for TagService
+     * Use when NOT injecting SessionFactoryService via Spring
+     * @param sessionFactoryService - instance of SessionFactoryService to use for initialization
+     */
+    public TagService(SessionFactoryService sessionFactoryService) {
```
benvogler: If all of the services are going to have this then it should just be done in the `ModelService` itself

src/main/java/com/monumental/services/ModelService.java
benvogler: Check out hibernate `Criteria` objects https://docs.jboss.org/hibernate/orm/3.5/javadocs/org/hibernate/Criteria.html

Instead of having this function be limited to one equals, you could instead have `ModelService.createCriteria`, then you do all your `.add`s to it and pass the completed `Criteria` object to a `ModelService.executeQuery(Criteria criteria)`

src/main/java/util/InitialDataLoadApplication.java
benvogler: You could override `doInsert` to achieve this pretty easily for one service

src/main/java/com/monumental/models/Model.java
njd9419: I added the `Identity` one. I was running into the issue that if I called the `validate` method and the `id` field was in the `NewOrExisting` group I would get a `ConstraintViolationException`.

src/main/java/util/InitialDataLoadApplication.java
njd9419: Don't feel like doing that in this PR, but we should in the future

src/main/java/com/monumental/models/Monument.java
njd9419: That's just what I found to work for the initial data load.

I'll increase it to 2048.

src/main/java/com/monumental/models/Monument.java
njd9419: It's used in the `InitialDataLoadApplication`, but they're probably not needed right? The validation should run on insert/update?

src/main/java/com/monumental/models/Monument.java
njd9419: Well, asserting that they aren't null on insert/update of a Monument would be covered by `@NotNull`, but I think we still need a way to associate the Contribution with the Monument in-memory before the insert/update like we do for Tags, no?

src/main/java/com/monumental/models/Model.java
benvogler: Shouldn't `Existing` work? That's how we had it in 343

src/main/java/com/monumental/models/Monument.java
benvogler: It seems like the entire point of these `addBlank` functions is to do a null check for you. Why not just initialize those lists in the `Monument` constructor if they're `null`?

```java
public Monument(...) {
    if (this.getContributions() == null) this.setContributions(new ArrayList<>());
    // etc.
}
```

src/main/java/com/monumental/models/Monument.java
benvogler: I'm not sure. They run on the controller methods so maybe not. But maybe there is a way to use `@Valid` outside of controller method params.

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -282,33 +301,41 @@ public void delete(List<Integer> ids) throws HibernateException {
...
-     * @return List<T> - the List of T returned by the query
+     * Perform a get query with the specified criteria
+     * @param criteria - List of Criterion objects to apply to the query
+     * @param initializeLazyLoadedCollections - If true, loads all of the collections associated with T that are
+     *                                        normally lazy loaded
+     * @return List<T> - List of T returned by the query
      */
     @SuppressWarnings("unchecked")
-    public List<T> getFromWhere(String whereParameterName, Object whereParameter) {
+    public List<T> getWithCriteria(List<Criterion> criteria, boolean initializeLazyLoadedCollections) {
```
benvogler: You should also add a `getWithCriteria(List<Criterion criteria> criteria)` that calls this method with `initializeLazyLoadedCollections` to false so that it's essentially an optional param. See the other get methods for reference

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -282,33 +301,41 @@ public void delete(List<Integer> ids) throws HibernateException {
...
         try {
             transaction = session.beginTransaction();
-            records = session.createQuery("FROM " + tableName + " WHERE " + whereParameterName + " = :whereParameter")
-                    .setParameter("whereParameter", whereParameter)
-                    .list();
+            Criteria c = session.createCriteria(this.getModelClass());
+
+            for (Criterion criterion : criteria) {
+                c.add(criterion);
+            }
```
benvogler: beauty

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -14,20 +14,8 @@
...
     public MonumentService(SessionFactoryService sessionFactoryService) {
-        this.sessionFactoryService = sessionFactoryService;
-    }
-
-    /**
-     * Public default constructor for MonumentService
-     */
-    public MonumentService() {
-
+        super(sessionFactoryService);
```
benvogler: Why do we even override the constructor here if it just calls super and returns?

src/main/java/com/monumental/services/TagService.java
```diff
@@ -35,9 +26,13 @@ public TagService() {
...
+     * @param initializeLazyLoadedCollections - If true, initializes all of the Monuments associated with the Tags
      * @return List<Tag> - List of Tags with the specified name
      */
-    public List<Tag> getTagsByName(String name) {
-        return this.getFromWhere("name", name);
+    public List<Tag> getTagsByName(String name, boolean initializeLazyLoadedCollections) {
+        ArrayList<Criterion> criteria = new ArrayList<>();
+        criteria.add(Restrictions.eq("name", name));
+
+        return this.getWithCriteria(criteria, initializeLazyLoadedCollections);
```
benvogler: This is really smoove. There might be some other `get`s out there we can replace with this but I don't remember

src/main/java/com/monumental/services/ModelService.java
benvogler: Also this is super nitpicky but I kind of prefer that the method just be called `get` and accept the criteria param. I don't know why but I like them all having the same name

src/main/java/com/monumental/models/Monument.java
njd9419: Fair enough you have provided all valid points. I'll fix that now.

src/main/java/com/monumental/services/ModelService.java
njd9419: I had to Google a couple of times which word was the plural and which was the singular.

src/main/java/com/monumental/services/MonumentService.java
njd9419: Model is an abstract class so you can't do `MonumentService monumentService = new ModelService(sessionFactoryService)`. How else would you do this?

src/main/java/com/monumental/models/Model.java
njd9419: We talked on Slack:

Essentially the Id field should use `Existing` and `NewOrExisting` on most of the other columns. Then when validating (either in a Controller method parameter or manually), use `New` or `Existing`.

The only difference between `Existing` and `New` will be the ID.

Fixing this in the coming commit.

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -20,6 +22,11 @@ public Monument createMonument(@RequestBody Monument monument) {
         return monument;
     }
 
+    @PostMapping("/api/monument/testValidation")
```
njd9419: This is just for testing and should be removed before merging.

njd9419:
As far as I know this is good to go.

@benvogler are the changes you requested done?

PR #33, opened 2019-10-19
Create Application to Generate Descriptions for Monument Data

njd9419:
This PR adds an application class that we can run to pull all of the Monument data currently in the database and formats it in a CSV file.

As part of this, I re-did how the `CsvFileReader` and `CsvFileWriter` work to make them a bit more usable.

Note that this PR was branched off of #32, so it has those changes in there as well. Once that's merged I can re-base onto master and fix any conflicts.

### DIFF
--- a/src/main/java/util/csvparsing/CsvFileHelper.java
+++ b/src/main/java/util/csvparsing/CsvFileHelper.java
@@ -1,12 +1,12 @@
 package util.csvparsing;
 
 /**
- * Class that defines common methods needed to interface with CSV Files
+ * Class that defines common methods needed when interfacing with CSV files
  */
-class CsvFileInterface {
+class CsvFileHelper {
 
     /**
-     * Method to determine if the filePath points to a CSV file
+     * Static method to determine if the filePath points to a CSV file
      * @param filePath - the path to the file to check, as a String
      * @return boolean - true if the file extension of filePath is ".csv", false otherwise
      */
--- a/src/main/java/util/csvparsing/CsvFileInterface.java
+++ b/src/main/java/util/csvparsing/CsvFileInterface.java
@@ -1,0 +1,19 @@
+package util.csvparsing;
+
+/**
+ * Class that defines common methods needed to interface with CSV Files
+ */
+class CsvFileInterface {
+
+    /**
+     * Method to determine if the filePath points to a CSV file
+     * @param filePath - the path to the file to check, as a String
+     * @return boolean - true if the file extension of filePath is ".csv", false otherwise
+     */
+    static boolean isCsvFile(String filePath) {
+        String[] fileExtensionArray = filePath.split("\\.");
+        String fileExtension = fileExtensionArray[fileExtensionArray.length - 1];
+
+        return fileExtension.equals("csv");
+    }
+}
--- a/src/main/java/util/csvparsing/CsvMonumentConverter.java
+++ b/src/main/java/util/csvparsing/CsvMonumentConverter.java
@@ -7,6 +7,7 @@ import com.monumental.models.Tag;
 
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
+import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
 
@@ -54,7 +55,7 @@ public class CsvMonumentConverter {
                     break;
                 case 3: // Date
                     // I made a couple of assumptions about the format of the dates in the file:
-                    // 1. The dates would be in the format dd-mm-yyyy
+                    // 1. The dates would be in the format dd-MM-yyyy
                     // 2. If the day and month were unknown, the cell would just contain yyyy
                     // 3. In the case described in 2, the day and month would be set to 01-01
                     // TODO: Agree upon a standard format for dates in CSV file uploads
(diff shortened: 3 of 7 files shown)


src/main/java/util/csvparsing/CsvFileInterface.java
```diff
@@ -0,0 +1,19 @@
+package util.csvparsing;
+
+/**
+ * Class that defines common methods needed to interface with CSV Files
+ */
+class CsvFileInterface {
```
njd9419: If you have a better name for this class, please let me know lol

src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -104,4 +361,27 @@ private static String removeBeginningAndEndingQuotes(String string) {
...
+        return "\"" + string + "\"";
+    }
+
+    /**
+     * Static method to get the header row for CSV file outputs
+     * @return String - the CSV header row
+     */
+    public static String getHeaderRow() {
+        return "ID,Title,Artist,Date,Material,Latitude,Longitude,City,State,Address,Inscription,Description," +
+                "Tags,Contributors,References";
```
njd9419: This was just the order I decided on for the purposes of this quick task. When we come back to this as part of the export to CSV requirement we can determine a new order.

src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -60,36 +74,279 @@ else if (dateArray.length == 3) {
...
+                    monument.addReference(newReference);
             }
         }
 
+        result.setMonument(monument);
+
         return result;
     }
 
+    /**
```
njd9419: Instead of just pulling the Monument's description, I decided it would be worth the effort to get a baseline for converting it into an entire CSV row with all of its attributes so we can build on it later when we need to support Monument CSV exports.

src/main/java/util/csvparsing/CsvFileInterface.java
benvogler: Steve.java

Seriously though, I wouldn't use Interface in the name if it's a class because it's a little confusing. Maybe `CsvFileHelper` and just instantiate it in the writer/reader instead of extending it

src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -60,36 +74,279 @@ else if (dateArray.length == 3) {
...
+
+        // Column 5: Material
+        String material = monument.getMaterial();
+        if (material != null && !material.isEmpty()) {
+            material = material.replace("\"", "");
+            if (material.contains(",")) {
+                csvRow += addBeginningAndEndingQuotes(material);
+            }
+            else {
+                csvRow += material;
```
benvogler: This same logic is repeated a lot, turn it into a function

benvogler:
Looks like your code from the other PR is in here as well, gonna wait until that's merged before I review so I'm not reviewing stuff twice

src/main/java/util/csvparsing/CsvFileHelper.java
```diff
@@ -0,0 +1,19 @@
+package util.csvparsing;
```
njd9419: Renamed this to be `CsvFileHelper`. No need to instantiate it right now as the only method it has is `static`.

src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -137,6 +142,91 @@ else if (dateArray.length == 3) {
         return result;
     }
 
+    /**
+     * Static method for converting a specified Monument into a CSV row
+     * @param monument - Monument to convert into a CSV row
+     * @return String - CSV row representing the specified Monument
+     */
+    public static String convertMonument(Monument monument) {
```
njd9419: Refactored this method to use some private helper methods. @benvogler I re-ran the GenerateMonumentDescriptionsApplication and everything seemed to match the original CSV generated, but double checking that the logic is still the same from the earlier commit would be appreciated.

benvogler:
I'll re-review after our meeting today

src/main/java/util/csvparsing/CsvMonumentConverter.java
benvogler: This is also useful because eventually I will pull all the addresses from google and will need to export that data for everyone else to use 👍

src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -137,6 +142,91 @@ else if (dateArray.length == 3) {
...
+
+        // Column 13: Tags
+        csvRow += convertTagsToCsvCell(monument.getTags());
+
+        csvRow += ",";
+
+        // Column 14: Contributors
+        csvRow += convertContributionsToCsvCell(monument.getContributions());
+
+        csvRow += ",";
```
benvogler: I would recommend `String.join(",", listOfCells)` although I realize this code is pretty temporary

njd9419: No :)

PR #35, opened 2019-10-20
Search bar

bxsmith007:
- Created Initial Search Bar
- Minor change to spacing on Links

### DIFF
--- a/client/src/components/Header/Header.js
+++ b/client/src/components/Header/Header.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import './Header.scss';
 import { NavLink } from 'react-router-dom';
-import { Button } from 'react-bootstrap';
+import { Collapse, Button, Form, FormControl, Nav, Navbar, NavItem } from 'react-bootstrap';
 
 export default class Header extends React.Component {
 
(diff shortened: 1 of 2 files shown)


client/src/components/Header/Header.js
```diff
@@ -17,12 +17,16 @@ export default class Header extends React.Component {
...
                         {this.links.map((link, index) =>
-                            <NavLink to={link.route} exact={link.exact} className="header-link" activeClassName="active" key={link.name}>{link.name}</NavLink>
+                            <NavLink to={link.route} exact={link.exact} className="header-link mr-4" activeClassName="active" key={link.name}>{link.name}</NavLink>
                         )}
                     </div>
                 </div>
                 <div className="center">
-                    Put a search bar here pls
+                    <Form inline>
+                        <FormControl type="text" placeholder="Title" size="sm" className="mr-sm-2" />
```
njd9419: I think we wanted to make it more obvious to the user that they could search by more than just Title in this search bar.

Maybe change the placeholder text to something like "Search by title, artist, material, tag..."

client/src/components/Header/Header.js
```diff
@@ -17,12 +17,16 @@ export default class Header extends React.Component {
...
+                        <FormControl type="text" placeholder="Location" size="sm" className="mr-sm-2" />
```
njd9419: Maybe the placeholder for this search bar could be "Near..." or something along those lines.

PR #40, opened 2019-10-22
Set First Image Shown on View Page to Primary Image

njd9419:
This PR sets the first image shown on the View Page to the primary Image rather than the first Image.

### DIFF
--- a/client/src/components/Gallery/Gallery.js
+++ b/client/src/components/Gallery/Gallery.js
@@ -27,11 +27,9 @@ export default class Gallery extends React.Component {
 
     render() {
         const images = JSON.parse(JSON.stringify(this.props.images));
-        let primaryImage = images.find(e => e.isPrimary === true);
+        let primaryImage = images.find(e => e.isPrimary);
 
-        primaryImage = primaryImage !== undefined ?
-            primaryImage :
-            images.splice(0, 1);
+        primaryImage = primaryImage || images.splice(0, 1)[0];
 
         if (!images || !images.length) {
             return (<div style={{marginBottom: '0'}}/>);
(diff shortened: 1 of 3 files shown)


client/src/components/Gallery/Gallery.js
```diff
@@ -27,13 +26,17 @@ export default class Gallery extends React.Component {
     }
 
     render() {
-        const images = JSON.parse(JSON.stringify(this.state.images));
+        const images = JSON.parse(JSON.stringify(this.props.images));
+        let primaryImage = images.find(e => e.isPrimary === true);
```
benvogler:
```suggestion
        let primaryImage = images.find(e => e.isPrimary);
```

client/src/components/Gallery/Gallery.js
```diff
@@ -27,13 +26,17 @@ export default class Gallery extends React.Component {
...
+
+        primaryImage = primaryImage !== undefined ?
+            primaryImage :
+            images.splice(0, 1);
```
benvogler: You can just do
```javascript
primaryImage = primaryImage || images.splice(0, 1)[0];
```
The reason being that JS will assign primaryImage to primaryImage if it's truthy (not undefined, null, 0, or false), or if it's falsy it will continue to the other side of the or and assign whatever the return value of that is

Also `splice` returns an array even if it's one element, which is why I had `[ primaryImage ] = images.splice(0, 1);` which is array destructuring and assigns the first index of the array to primaryImage

njd9419: Thank you for your wisdom kind wizard.

client/src/components/Gallery/Gallery.js
njd9419: Done.

PR #41, opened 2019-10-22
Upgrade hibernate to version 5, use JPA CriteriaQuery for searching

benvogler:
While researching some ways to elegantly build the search query (since it will be rather complex due to lots of optional filters) I realized we were behind by a major version in hibernate which actually changed a lot of hibernate including the criteria querying we had just done in another PR.

This PR upgrades hibernate to 5 and uses CriteriaQuery in the SearchController to test out the criteria building

### DIFF
--- a/src/main/java/com/monumental/ExceptionHandler.java
+++ b/src/main/java/com/monumental/ExceptionHandler.java
@@ -1,0 +1,22 @@
+package com.monumental;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@ControllerAdvice
+public class ExceptionHandler {
+
+    @org.springframework.web.bind.annotation.ExceptionHandler(Exception.class)
+    public ResponseEntity<Object> handle(Exception ex,
+                                         HttpServletRequest request, HttpServletResponse response) {
+        ex.printStackTrace();
+        if (ex instanceof NullPointerException) {
+            return ResponseEntity.badRequest().body(ex);
+        }
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(ex);
+    }
+}
--- a/src/main/java/com/monumental/ExceptionHandler.java
+++ b/src/main/java/com/monumental/ExceptionHandler.java
@@ -7,6 +7,11 @@ import org.springframework.web.bind.annotation.ControllerAdvice;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+/**
+ * This class is responsible for intercepting all exceptions and deciding what should happen with the HTTP response
+ * It also is able to log these exceptions, which is really helpful for debugging since Spring doesn't always print
+ * stacktraces for most exceptions
+ */
 @ControllerAdvice
 public class ExceptionHandler {
 
--- a/src/main/java/com/monumental/controllers/SearchController.java
+++ b/src/main/java/com/monumental/controllers/SearchController.java
@@ -7,7 +7,6 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
-import javax.transaction.Transactional;
 import java.util.List;
 
 @RestController
@@ -23,7 +22,6 @@ public class SearchController {
      * @return            Matching Monuments based on their title
      */
     @GetMapping("/api/search")
-    @Transactional
     public List<Monument> searchMonuments(@RequestParam(required = false, value = "q") String searchQuery,
                                           @RequestParam(required = false, defaultValue = "1") String page,
                                           @RequestParam(required = false, defaultValue = "25") String limit) {
--- a/src/main/java/com/monumental/controllers/SearchController.java
+++ b/src/main/java/com/monumental/controllers/SearchController.java
@@ -28,32 +28,22 @@ public class SearchController {
      */
     @GetMapping("/api/search")
     public List<Monument> searchMonuments(@RequestParam(required = false, value = "q") String searchQuery,
-                                          @RequestParam(required = false, defaultValue = "1", value = "page") String pageString,
-                                          @RequestParam(required = false, defaultValue = "25", value = "limit") String limitString) {
+                                          @RequestParam(required = false, defaultValue = "1") String page,
+                                          @RequestParam(required = false, defaultValue = "25") String limit) {
 
-        Integer page = null;
-        if (pageString != null) page = Integer.parseInt(pageString);
-        Integer limit = null;
-        if (limitString != null) limit = Integer.parseInt(limitString);
+        CriteriaBuilder builder = monumentService.getCriteriaBuilder();
+        CriteriaQuery<Monument> query = monumentService.createCriteriaQuery(builder, false);
+        Root<Monument> root = monumentService.createRoot(query);
 
-        try {
-            CriteriaBuilder builder = monumentService.getCriteriaBuilder();
-            CriteriaQuery<Monument> query = monumentService.createCriteriaQuery(builder, false);
-            Root<Monument> root = monumentService.createRoot(query);
-
-            if (searchQuery != null) {
-                query.where(
-                    builder.or(
-                        builder.equal(builder.function("fts", Boolean.class, root.get("title"), builder.literal(searchQuery)), true),
-                        builder.equal(builder.function("fts", Boolean.class, root.get("artist"), builder.literal(searchQuery)), true)
-                    )
-                );
-            }
-
-            return this.monumentService.getWithCriteriaQuery(query);
-        } catch (Exception e) {
-            e.printStackTrace();
+        if (searchQuery != null) {
+            query.where(
+                builder.or(
+                    builder.equal(builder.function("fts", Boolean.class, root.get("title"), builder.literal(searchQuery)), true),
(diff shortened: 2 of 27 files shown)


src/main/java/com/monumental/controllers/SearchController.java
```diff
@@ -20,11 +23,37 @@
...
+            if (searchQuery != null) {
+                query.where(
+                    builder.or(
+                        builder.equal(builder.function("fts", Boolean.class, root.get("title"), builder.literal(searchQuery)), true),
+                        builder.equal(builder.function("fts", Boolean.class, root.get("artist"), builder.literal(searchQuery)), true)
+                    )
+                );
+            }
+
+            return this.monumentService.getWithCriteriaQuery(query);
```
benvogler: This is the usage of the new CriteriaBuilder stuff, which allows us to:
1. Call custom functions (in this case, "fts" which is doing the Full Text Search)
2. Build complex logic with `builder.or`, `builder.and`, etc.

src/main/java/com/monumental/controllers/SearchController.java
```diff
@@ -20,11 +23,37 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        return null;
```
benvogler: I'll get rid of this in the next PR

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -397,4 +407,36 @@ private void initializeLazyLoadedCollections(List<T> records) {
...
+        CriteriaQuery<T> query = builder.createQuery(this.getModelClass());
+        if (setRoot) this.createRoot(query);
+        return query;
+    }
+
+    public Root<T> createRoot(CriteriaQuery<T> query) {
+        Root<T> root = query.from(this.getModelClass());
+        query.select(root);
+        return root;
+    }
```
benvogler: These are basically all provided to make it easier to use the `CriteriaBuilder`, so that you don't necessarily have to do every step of the process every time you want to use it, but if you need any of the intermediary objects (such as `Root`) you are able to do so and then specify it to the later steps.

Usually you should just be able to call `createCriteriaQuery()`, set your clauses, and query it.

src/main/resources/application.properties
```diff
@@ -5,4 +5,16 @@ spring.http.encoding.enabled=true
...
+spring.datasource.driver-class-name=org.postgresql.Driver
+spring.datasource.username=postgres
+spring.datasource.password=password
+spring.datasource.url=jdbc:postgresql://localhost:5432/postgres
+spring.jpa.properties.hibernate.dialect=com.monumental.CustomPostgreSQL9Dialect
+spring.jpa.properties.hibernate.connection.driver_class=org.postgresql.Driver
+spring.jpa.properties.hibernate.jdbc.lob.non_contextual_creation=true
+spring.jpa.show-sql = true
+spring.jpa.hibernate.ddl-auto = update
+spring.jpa.properties.hibernate.default_schema=public
```
benvogler: I *think* that this completely replaces hibernate.cfg.xml but I have not tested it fully

benvogler:
I'm noticing a problem where after many refreshes I stop getting results back until I restart the server. I think maybe this is not closing the DB connection properly. I'll investigate this this week, for now I'm marking this as do not merge

benvogler:
The issue I mentioned above was caused by the references being used in the description generation. This seems to be causing some update loop or something. So for now, I've just removed references from the description and updated the related tests. We'll need to figure out if there's a way around that

src/main/java/com/monumental/ExceptionHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * This class is responsible for intercepting all exceptions and deciding what should happen with the HTTP response
+ * It also is able to log these exceptions, which is really helpful for debugging since Spring doesn't always print
+ * stacktraces for most exceptions
+ */
+@ControllerAdvice
```
njd9419: What does '@ControllerAdvice' do?

src/main/java/com/monumental/controllers/SearchController.java
```diff
@@ -20,11 +20,22 @@
      * This function lets you search monuments using the q query param
      * Ex: GET http://localhost:8080/api/search?q=Memorial
```
njd9419: This comment should be updated with the new request parameters.

src/main/java/com/monumental/controllers/SearchController.java
```diff
@@ -20,11 +20,22 @@
...
      * TODO: Possibly search related tables such as Tags
-     * @param q The search query string
-     * @return  Matching Monuments based on their title
+     * @param searchQuery The search query string
+     * @return            Matching Monuments based on their title
      */
     @GetMapping("/api/search")
-    public List<Monument> searchMonuments(@RequestParam String q) {
-        return this.monumentService.search(q);
+    @Transactional
```
njd9419: Can you explain what the `@Transactional` annotation does?

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -397,4 +421,62 @@ private void initializeLazyLoadedCollections(List<T> records) {
...
+     * TODO: Delete and replace with JPA 2 code
+     */
+    Session openSession() {
+        return this.sessionFactoryService.getFactory().openSession();
+    }
+
+    /**
+     * This is handles your connection to the database and has the CriteriaBuilder reference
+     */
+    public EntityManager getEntityManager() {
```
njd9419: Any reason this is in a function and we're not just calling `this.entityManager.blah`?

src/main/java/com/monumental/services/ModelService.java
```diff
@@ -397,4 +421,62 @@ private void initializeLazyLoadedCollections(List<T> records) {
...
+
+    /**
+     * Creates a CriteriaQuery and optionally creates its Root for you and sets it to distinct mode (no duplicates)
+     * @param builder   Your CriteriaBuilder, from getCriteriaBuilder()
+     * @param setRoot   If true, a Root will be created for you and set on the CriteriaQuery
+     *                  You may choose to leave this false if you need a reference to the Root in order to do things
+     *                  such as root.fetch("collectionName", FetchType.LEFT) if you want to include a lazy-loaded
+     *                  collection in your query results
+     */
+    public CriteriaQuery<T> createCriteriaQuery(CriteriaBuilder builder, Boolean setRoot) {
```
njd9419: Any reason this is a `Boolean` object and not the primitive?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -19,34 +17,59 @@ public MonumentService(SessionFactoryService sessionFactoryService) {
...
+    /**
+     * Count the total number of results for a FTS search
+     * TODO: Be sure to update this whenever the search function changes so that they stay in sync
+     */
+    public Integer countSearchResults(String searchQuery) {
+
+        CriteriaBuilder builder = this.getCriteriaBuilder();
+        CriteriaQuery<Long> query = builder.createQuery(Long.class);
+        Root<Monument> root = query.from(Monument.class);
+        query.select(builder.countDistinct(root)).distinct(true);
```
njd9419: Is doing `builder.countDistinct(root)).distinct(true)` necessary? Is this double the distinct-ness?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -19,34 +17,59 @@ public MonumentService(SessionFactoryService sessionFactoryService) {
...
+        List<Monument> monuments = this.getWithCriteriaQuery(query, Integer.parseInt(limit), (Integer.parseInt(page)) - 1);
+
+        return monuments;
+    }
+
+    /**
+     * Count the total number of results for a FTS search
+     * TODO: Be sure to update this whenever the search function changes so that they stay in sync
+     */
+    public Integer countSearchResults(String searchQuery) {
```
njd9419: Where is this used? Is there missing front-end code that uses this? (I'm assuming so).

njd9419: Sorry, I meant the Controller method that calls this, but same thing.

src/main/java/com/monumental/ExceptionHandler.java
benvogler: I think it just means it gets applied to all the controllers. It's the standard way to do exception handling in Spring
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/ControllerAdvice.html

src/main/java/com/monumental/services/MonumentService.java
benvogler: Yes it's used by the search page. I wanted to try to split back & front end pieces so that the PRs are a little more manageable

src/main/java/com/monumental/services/MonumentService.java
benvogler: I was losing my mind at the time trying to figure out why my count was always off. No, it's not necessary.

Turned out it was because in the search query I was using `.fetch("tags")` to load in the tags, but this was actually making it so only monuments with tags were returned while the count was returning those without tags too. The reason was because I had to set the join type of the fetch to `JoinType.LEFT` so that monuments without tags were returned

src/main/java/com/monumental/services/ModelService.java
benvogler: I don't think you can make Spring beans public. Also, muh getters and setters. Technically this makes it readonly to the public

src/main/java/com/monumental/controllers/SearchController.java
benvogler: I think it makes the session stay open for the duration of the function, which I thought I needed to get the tags but because I'm using `.fetch` to get the monuments and their tags in one query it's not necessary so I removed it

benvogler:
I have to wait until this is merged to PR the front end because it is branched off of this

PR #42, opened 2019-10-23
Added Inscription to View M&M Page

njd9419:
This PR adds the Inscription field to the View M&M Page.

### DIFF
--- a/client/src/components/Monument/Monument.js
+++ b/client/src/components/Monument/Monument.js
@@ -83,13 +83,20 @@ export default class Monument extends React.Component {
     capitalize(word) {
         if (word) return word.charAt(0).toUpperCase() + word.slice(1).trim();
         else return word;
-    }
+    };
 
     parseState(state) {
         if (!state) return state;
         if (state.toLowerCase() === 'dc') state = 'd.c.';
         return state.toUpperCase().trim();
     };
+
+    formatInscription(inscription) {
+        if (!inscription) return inscription;
+        // This nifty little regex removes all of the double quotes in the string
+        inscription = inscription.replace(/["]+/g, '');
+        return '"' + inscription + '"';
+    }
 
     render() {
         if (this.state.error) return this.renderError();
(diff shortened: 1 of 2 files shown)


client/src/components/Monument/Monument.js
```diff
@@ -83,14 +83,21 @@ export default class Monument extends React.Component {
     capitalize(word) {
         if (word) return word.charAt(0).toUpperCase() + word.slice(1).trim();
         else return word;
-    }
+    };
```
benvogler:
```suggestion
    }
```
I think there's semicolons on some of the class functions here because I copied them from another place, you usually don't put semicolons on class functions in JS

client/src/components/Monument/Monument.js
```diff
@@ -255,6 +263,25 @@ export default class Monument extends React.Component {
...
     }
 
+    renderInscription() {
+        const { monument } = this.state;
+
+        let inscription;
+        if (monument.inscription) {
+            inscription = (
+                <div>
+                    <b>Inscription:</b> {this.formatInscription(monument.inscription)}
```
benvogler: Don't use `<b>` it's deprecated (although there's `<strong>` now which is apparently ok to use), you should use CSS for bold: `font-weight: bold`, although bootstrap provides some convenient classes that you can use i.e.
```html
<span className="font-weight-bold">Inscription</span>
```

client/src/components/Monument/Monument.js
benvogler: also I will get rid of those other ones if I ever remember

benvogler: feel free to skip this suggestion for now, super minor

njd9419: I'll remove them all

client/src/components/Monument/Monument.js
njd9419: I shall do that.

Can you explain what a `<span>` tag does in a very noob way? :)

PR #43, opened 2019-10-23
Update image gallery appearance

benvogler:
I didn't like the image gallery so I made it look more typical...

Downside of this approach is that it doesn't really work if you have >16 images... don't think that's expected at this point though 😄 

The image automatically rotates every 10 seconds and has a nice animation
image

I also changed the modal to look less modally and more image previewy
image

### DIFF
--- a/client/src/components/Gallery/Gallery.js
+++ b/client/src/components/Gallery/Gallery.js
@@ -14,7 +14,7 @@ export default class Gallery extends React.Component {
         const imageRotationInterval = (props.imageRotationInterval || 10000) - imageRotationAnimationLength;
         this.state = {
             modalOpen: false,
-            modalPage: 0,
+            modalImageIndex: 0,
             selectedImageIndex,
             imageRotationInterval,
             imageRotationAnimationLength,
@@ -27,17 +27,15 @@ export default class Gallery extends React.Component {
         let index = this.state.selectedImageIndex;
         index++;
         if (index === this.props.images.length) index = 0;
-        const previousIndex = this.state.selectedImageIndex;
-        this.animate(index, previousIndex);
+        this.animate(index);
     }
 
     selectImage(index) {
         window.clearTimeout(this.state.imageTimeout);
-        const previousIndex = this.state.selectedImageIndex;
-        this.animate(index, previousIndex);
+        this.animate(index);
     }
 
-    animate(index, previousIndex) {
+    animate(index) {
         this.setState({
             animationIndex: index,
             animating: true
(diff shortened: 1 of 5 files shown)


client/src/components/Gallery/Gallery.js
```diff
@@ -1,17 +1,59 @@
...
+        this.animate(index, previousIndex);
+    }
+
+    animate(index, previousIndex) {
+        this.setState({
+            animationIndex: index,
+            animating: true
+        });
+        const { imageTimeout, imageRotationAnimationLength } = this.state;
+        window.setTimeout(() => {
```
njd9419: Can you explain what this fancy code is doing?

client/src/components/Gallery/Gallery.js
```diff
@@ -58,109 +104,60 @@ export default class Gallery extends React.Component {
...
+                                  onClick={() => {
+                                      if (this.state.animating) return;
+                                      this.selectImage(index);
+                                  }}/>
+                        )
+                    })
+                }
             </div>
-        )
+        );
```
njd9419: Is it normal to have semi-colons after after `return` statements? I think we're inconsistent with that.

client/src/components/Monument/Monument.js
```diff
@@ -35,13 +35,22 @@ export default class Monument extends React.Component {
             this.callEndpoint(`/api/tags/?monumentId=${monumentId}`)
         ]).catch(err => error = err);
 
+        images.splice(0);
```
njd9419: Debug code?

client/src/components/Monument/Monument.js
```diff
@@ -35,13 +35,22 @@ export default class Monument extends React.Component {
...
+
         if (error) {
             console.error(error);
             this.setState({error: error});
             return;
         }
 
-        this.setState({monument, contributions, images, references, tags});
+        this.setState({monument, contributions, images: [
```
njd9419: This is good for now, but we'll definitely want to remove this once we get images stored in the back-end.

client/src/components/Gallery/Gallery.js
benvogler: I can try: basically, there's 2 image divs, one sits on top of the other covering it. The bottom div is our "real" image, the top one is the one used solely by the animation. The way the animation works is basically:

No animation is running
Image A is in bottom div
Top div is empty and not showing

Then, we change to Image B and this happens:
Animation is running
Image B is placed in bottom div
Image A is placed in top div with an opacity of 1, slowly animating to opacity 0, in other words disappearing

Animation is complete
Image B is still in the bottom div
The top div is now made fully invisible (due to its black background, it can't be visible while not animating)

The javascript that makes this work basically runs on a 1 second timer which is what setTimeout does. There are a lot of nuances to it that are easier to explain in person.

client/src/components/Gallery/Gallery.js
benvogler: Is it normal to have semi-colons after `return` statements in JS: Yes
Is it normal to have semi-colons after `return` statements in React that return JSX (The JS/HTML stuff wrapped in `()`): I'm not sure

But since the first one is true, I tend to do that when I remember although it's easy to forget with these massive blocks of JSX in between sometimes

client/src/components/Monument/Monument.js
benvogler: derp

client/src/components/Monument/Monument.js
benvogler: also derp. although it's handy to have so I'll leave it in and put a todo

client/src/components/Gallery/Gallery.js
benvogler: I also just added even more complexity to this by delaying part of the last step 100ms to give the image time to load in, otherwise you would get a weird flashing glitch as the image popped into a new div

PR #45, opened 2019-10-25
Use similarity search instead of FTS, and order by most similar

benvogler:
I found that the pg_trgm plugin has a function `similarity` that takes two strings and returns a number between 0 and 1 that is the score of how similar the two strings are. Using this, we can find results that are within a certain "threshold" of similarity to our search string.

I'm using 0.1 currently because it seemed to work fine and I don't think we have any need to be more strict than that, especially since this now also orders the results by the sum of the similarity to the title and artist. This is inherently biased towards monuments with short titles, due to the way the similarity function works. I don't think there's any way to fix that unfortunately.

NOTE: In order for this to work you must enable the pg_trgm plugin. It comes with Postgres 9.6 but is not on by default. You can do this in pgAdmin:

Then just search for pg_trgm in the name field and press save. I will enable this on the production server later.

Here's some examples of it in action:

When searching for "Vietnam":


src/main/java/com/monumental/CustomPostgreSQL9Dialect.java
```diff
@@ -9,6 +9,6 @@
 public class CustomPostgreSQL9Dialect extends PostgreSQL9Dialect {
 
     public CustomPostgreSQL9Dialect() {
-        registerFunction("fts", new FTSFunction());
+        // TODO: Either use this for location searching or delete it
```
benvogler: @njd9419 I left this class in, in case you need it for location searching. We can remove it if you end up not needing it.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -42,19 +53,17 @@ private void buildFTSQuery(CriteriaBuilder builder, CriteriaQuery query, Root ro
         CriteriaQuery<Monument> query = this.createCriteriaQuery(builder, false);
         Root<Monument> root = this.createRoot(query);
         root.fetch("tags", JoinType.LEFT);
-        query.select(root).distinct(true);
```
benvogler: We can't use distinct because we're ordering by the functions which are not in the SELECT clause. That being said, I haven't noticed any duplicate results yet so I'm tentatively hoping that it isn't necessary anyway for this query

src/main/java/com/monumental/CustomPostgreSQL9Dialect.java
njd9419: Sounds good.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -42,19 +53,17 @@ private void buildFTSQuery(CriteriaBuilder builder, CriteriaQuery query, Root ro
         CriteriaQuery<Monument> query = this.createCriteriaQuery(builder, false);
```
njd9419: The comment above the `search` method should be fixed.

src/main/java/com/monumental/CustomPostgreSQL9Dialect.java
njd9419: Also, doesn't this mean we can change our `hibernate.dialect` in our configuration files?

benvogler: Yes although I left it as is since I wasn't sure if you would need this for PostGIS

njd9419: Thankfully PostGIS has its own dialect so we should be good.

src/main/java/com/monumental/services/MonumentService.java
njd9419: Please :)

benvogler: @njd9419

PR #46, opened 2019-10-26
First Pass - Location-Based Searching (Back-End)

njd9419:
This PR takes a first pass at implementing location-based searching on the back-end.

It makes use of PostGIS database-level functions to determine the distance between two `Point`s.

There is some manual setup required for PostGIS, I can give you guys a hand with that.

Some of this is subject to change after the front-end searching is completed.


build.gradle
```diff
@@ -40,12 +43,14 @@ dependencies {
     implementation('com.sun.xml.bind:jaxb-core:2.3.0')
     implementation('com.fasterxml.jackson.datatype:jackson-datatype-hibernate5:2.10.0')
     implementation group: 'org.hibernate', name: 'hibernate-search-orm', version: '5.11.3.Final'
+    implementation group: 'org.hibernate', name: 'hibernate-spatial', version: '5.1.0.Final'
     implementation group: 'org.postgresql', name: 'postgresql', version: '42.1.4'
     implementation('com.google.maps:google-maps-services:0.10.1')
+    implementation('com.bedatadriven:jackson-datatype-jts:2.4')
+    implementation group: 'org.opengeo', name: 'geodb', version: '0.8'
```
njd9419: This dependency includes H2, so I remove it from below.

src/main/java/com/monumental/config/AppConfig.java
```diff
@@ -32,4 +33,13 @@ public ResourceBundleMessageSource messageSource() {
...
+
+    /**
+     * This Bean is required so that the JSON deserializer knows how to deserialize GeoJson since it does not
+     * know how to by default
+     */
+    @Bean
+    public JtsModule jtsModule() {
+        return new JtsModule();
+    }
 }
```
njd9419: 🅱️ ean

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -44,40 +44,94 @@ private void buildSimilarityQuery(CriteriaBuilder builder, CriteriaQuery query,
...
+                                                builder.literal(4326)
+                                        ),
+                                        builder.literal(2877)
+                                ),
+                                builder.literal(feet)
+                        ),
+                true)
+        );
+    }
+
```
njd9419: `ST_DWithin` and `ST_Transform` are PostGIS database functions.

Don't even get me started about the `2877` and `4326`...

benvogler: can you make them `public static final int` with a link to one of those sites that explains them

src/main/java/com/monumental/config/AppConfig.java
benvogler: 🅱️ ean

build.gradle
benvogler: Maybe put a comment in here about that

src/main/java/com/monumental/controllers/SearchController.java
```diff
@@ -16,23 +16,35 @@
...
      */
     @GetMapping("/api/search")
     public List<Monument> searchMonuments(@RequestParam(required = false, value = "q") String searchQuery,
                                           @RequestParam(required = false, defaultValue = "1") String page,
-                                          @RequestParam(required = false, defaultValue = "25") String limit) {
-        return monumentService.search(searchQuery, page, limit);
+                                          @RequestParam(required = false, defaultValue = "25") String limit,
+                                          @RequestParam(required = false, value = "lat") String latitude,
+                                          @RequestParam(required = false, value = "lon") String longitude,
+                                          @RequestParam(required = false, defaultValue = "25") String miles) {
```
benvogler:
```suggestion
                                          @RequestParam(required = false, defaultValue = "25", value = "d") String distance) {
```

1. I like `d` because it's short and god knows these urls are going to be massive
2. I like distance over miles because it describes what this does better

src/main/java/com/monumental/models/Monument.java
```diff
@@ -41,11 +42,8 @@
     @NotNull(groups = {New.class, Existing.class}, message = "Material can not be null")
     private String material;
 
-    @Column(name = "lat")
-    private Double lat;
-
-    @Column(name = "lon")
-    private Double lon;
+    @Column(name = "point", columnDefinition = "geometry")
+    private Point point;
```
benvogler: [Point point Point, point point Point point, point point. Point. Point, point.](https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo)

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -44,40 +44,94 @@ private void buildSimilarityQuery(CriteriaBuilder builder, CriteriaQuery query,
...
         query.select(builder.countDistinct(root));
 
-        this.buildSimilarityQuery(builder, query, root, searchQuery, 0.1, false);
+        if (searchQuery != null) {
+            this.buildSimilarityQuery(builder, query, root, searchQuery, 0.1, false);
+        }
+
+        if (latitude != null && longitude != null && miles != null) {
+            this.buildDWithinQuery(builder, query, root, latitude, longitude, Integer.parseInt(miles));
+        }
```
benvogler: I'm realizing now that this whole thing will always be a duplicate of what's in `search` so we should really abstract it to a function. Can be in a different PR though, like when we add filters

src/main/java/com/monumental/models/Monument.java
benvogler: Also this should probably be `Point coordinates` to be more descriptive

src/main/java/com/monumental/triggers/MonumentTrigger.java
```diff
@@ -17,7 +17,12 @@ void beforeInsert(Monument record) {
         if (record.getAddress() != null) {
             return;
         }
-        String address = googleMapsService.getAddressFromCoordinates(record.getLat(), record.getLon());
+
+        if (record.getPoint() == null) {
+            return;
+        }
+
+        String address = googleMapsService.getAddressFromCoordinates(record.getPoint().getY(), record.getPoint().getX());
```
benvogler: Might make sense to add `getLat` and `getLon` to `Monument` as a shortcut for these `getY` and `getX` functions

benvogler:
Looks good just a couple of minor things

src/main/java/com/monumental/services/MonumentService.java
njd9419: Agreed.

PR #48, opened 2019-10-31
Working full-page map and map showing search results on search results page

benvogler:


client/src/actions/map.js
```diff
@@ -0,0 +1,38 @@
...
+function fetchMonumentsError(error) {
+    return {
+        type: FETCH_MAP_MONUMENTS_ERROR,
+        error: error
+    };
+}
+
+/**
+ * Queries for all monuments, to be displayed on the monument map page
+ * This is an async action (redux-thunk)
```
njd9419: Gotta tell ya, redux does a whole lot of thinking

client/src/components/Monument/Map/Map.js
```diff
@@ -7,11 +7,15 @@ export default class Map extends React.Component {
...
+
         return (
             <div className="visit">
                 <div className="map">
                     <iframe title="gmaps-iframe"
-                            src={`https://maps.google.com/maps?q=${monument.address ? monument.address : monument.coordinatePointAsString}&z=16&output=embed`}
+                            src={`https://maps.google.com/maps?q=${q}&z=16&output=embed`}
                             frameBorder="0"/>
                 </div>
             </div>
```
njd9419: Our PRs are going to have conflicts on this. I'll remove how I did mine.

njd9419: Done.

client/src/components/Search/MapResults/MapResults.js
```diff
@@ -0,0 +1,43 @@
...
+                        </Popup>
+                    </CircleMarker>
+                ));
+            } else {
+                markers.push((
+                    <Marker key={monument.id} position={[monument.lat, monument.lon]}>
+                        <Popup>
+                            <a href={'/monuments/' + monument.id}>{monument.title}</a>
+                            <Address monument={monument}/>
+                        </Popup>
```
njd9419: Could we possibly pull `<Popup>` out of these so it's not duplicated?

client/src/components/Search/MapResults/MapResults.js
```diff
@@ -0,0 +1,43 @@
...
+                    </Marker>
+                ));
+            }
+        }
+        return (
+            <Map center={[39.8283, -98.5795]} zoom={zoom}>
```
njd9419: Maybe add a comment that this is the geographic center of the U.S.

client/src/components/Search/MapResults/MapResults.js
```diff
@@ -0,0 +1,43 @@
...
+                <TileLayer
```
njd9419: What does this component do?

client/src/components/Search/Search.scss
```diff
@@ -58,4 +64,13 @@
     .page-link {
         border: 0 !important;
     }
+}
+
+.map-results {
+    .leaflet-container {
+        position: sticky !important;
```
njd9419: What does `!important` do?

client/src/constants.js
```diff
@@ -15,3 +15,7 @@ export const FETCH_MONUMENT_ERROR = 'FETCH_MONUMENT_ERROR';
 export const SEARCH_MONUMENTS_PENDING = 'SEARCH_MONUMENTS_PENDING';
 export const SEARCH_MONUMENTS_SUCCESS = 'SEARCH_MONUMENTS_SUCCESS';
 export const SEARCH_MONUMENTS_ERROR = 'SEARCH_MONUMENTS_ERROR';
+
+export const FETCH_MAP_MONUMENTS_PENDING = 'FETCH_MAP_MONUMENTS_PENDING';
+export const FETCH_MAP_MONUMENTS_SUCCESS = 'FETCH_MAP_MONUMENT_SUCCESS';
```
njd9419: `FETCH_MAP_MONUMENT_SUCCESS` -> `FETCH_MAP_MONUMENTS_SUCCESS`

client/src/actions/map.js
benvogler: really makes you thunk

client/src/components/Search/MapResults/MapResults.js
benvogler: It tells it to load in the background map tiles basically

client/src/components/Search/Search.scss
benvogler: There's an explanation here
https://css-tricks.com/when-using-important-is-the-right-choice/

It is generally bad practice to use !important because it makes it very difficult to then override your styles from another place. However, I think it's ok here for two reasons:

1. The `.leaflet-container` is having its position set to `relative` with in-line styles, which gives it a high importance making it hard to override without using `!important`
2. This CSS will only be loaded on the Search component, so other components that use the `.leaflet-container` will not even be affected by this or have to attempt to override the `!important`

client/src/components/Search/MapResults/MapResults.js
benvogler: Fixed

client/src/components/Search/MapResults/MapResults.js
benvogler: Made it a constant

client/src/constants.js
benvogler: Fixed

PR #50, opened 2019-11-01
Add Description to Monument Text Search

njd9419:
This PR adds the ability to search by the Monument's Description field when doing a text search.

I also refactored the way we count the number of search results.


src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -25,28 +25,33 @@ public MonumentService(SessionFactoryService sessionFactoryService) {
...
-                builder.gt(builder.function("similarity", Number.class, root.get("artist"), builder.literal(searchQuery)), threshold)
+                builder.gt(builder.function("similarity", Number.class, root.get("artist"), builder.literal(searchQuery)), threshold),
+                builder.gt(builder.function("similarity", Number.class, root.get("description"), builder.literal(searchQuery)), threshold)
             )
         );
 
         if (orderByResults) {
             query.orderBy(
                 builder.desc(
                     builder.sum(
```
njd9419: `builder.sum()` only takes 2 arguments, so I had to use 2 sums.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -129,20 +133,6 @@ private void buildDWithinQuery(CriteriaBuilder builder, CriteriaQuery query, Roo
...
-        if (searchQuery != null) {
-            this.buildSimilarityQuery(builder, query, root, searchQuery, 0.1, false);
-        }
-
-        if (latitude != null && longitude != null && distance != null) {
-            this.buildDWithinQuery(builder, query, root, latitude, longitude, Integer.parseInt(distance));
-        }
-
-        return this.getEntityManager().createQuery(query).getSingleResult().intValue();
+        return this.search(searchQuery, "1", "25", latitude, longitude, distance).size();
```
benvogler: This is not correct. Size will be the paginated size. `countDistinct` is the size of ALL the results not just those in the page.

njd9419: I refactored how `search()` works so it can handle `null` values for `limit` and `page`.

benvogler: This is still a much slower query to make than a `count()` query. Whichever way we do it, we need to be using `countDistinct` for efficiency

njd9419: Oh my bad, I was confused what `countDistinct` was.

You're saying we should do something similar to: https://stackoverflow.com/questions/2883887/in-jpa-2-using-a-criteriaquery-how-to-count-results right?

benvogler: Yes which is basically what we had except with countDistinct instead of count

njd9419: I see.

I wanted to try to find a way to not have the search logic duplicated.

I'll fix it up and see if I can do that another way.

njd9419: Fixed it up. Should be good to go now.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -100,48 +129,33 @@ private void buildDWithinQuery(CriteriaBuilder builder, CriteriaQuery query, Roo
...
-            this.buildSimilarityQuery(builder, query, root, searchQuery, 0.1, true);
-        }
-
-        if (latitude != null && longitude != null && distance != null) {
-            this.buildDWithinQuery(builder, query, root, latitude, longitude, Integer.parseInt(distance));
-        }
+        this.buildSearchQuery(builder, query, root, searchQuery, latitude, longitude, distance);
 
-        List<Monument> monuments = this.getWithCriteriaQuery(query, Integer.parseInt(limit), (Integer.parseInt(page)) - 1);
+        List<Monument> monuments = limit != null
```
benvogler: Not sure I love the nested ternaries lol

njd9419: 😍

PR #51, opened 2019-11-03
S3 Image Integration

njd9419:
This PR updates the InitialDataLoadApplication to correctly save the S3 Object URL for the images associated with a Monument into the database.

By doing this, the front-end is now able to render the appropriate image on the View M&M page.

I manually uploaded all of the initial Monument images to a personal S3 bucket to make this possible. At some point we'll have to migrate them over to a bucket owned by the sponsor.


build.gradle
```diff
@@ -47,6 +47,7 @@ dependencies {
     implementation group: 'org.postgresql', name: 'postgresql', version: '42.1.4'
     implementation('com.google.maps:google-maps-services:0.10.1')
     implementation('com.bedatadriven:jackson-datatype-jts:2.4')
+    implementation('com.amazonaws:aws-java-sdk-s3:1.11.665')
```
njd9419: I added this for use in the `AwsS3Service` class (see comments below).

client/src/components/Monument/Gallery/Gallery.js
```diff
@@ -36,6 +36,12 @@ export default class Gallery extends React.Component {
     }
 
     animate(index) {
+        const { images } = this.props;
+        if (images.length <= 1) {
+            this.setState({animating: false});
+            return;
+        }
+
```
njd9419: The image animation was occurring even though there was only one image, which looked funky.

src/main/java/com/monumental/services/AwsS3Service.java
```diff
@@ -0,0 +1,72 @@
+package com.monumental.services;
```
njd9419: I realize that we might not have a need for SDK integration with S3 on the back-end, but I had already made this service class and figured it might be useful for something.

If it turns out it isn't, I can remove it.

njd9419: Now that I think about it more, this might be useful for bulk uploading with images.

benvogler: Bulk uploading is going to happen on the front end, same as single upload, and we would still end up uploading images twice (once to our server, then once from our server to s3) if we do it this way. I think it just doubles the upload time for no good reason when we could just simply upload to s3 from the client (which AWS provides a secure way to do).

src/main/java/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -142,6 +140,17 @@ else if (dateArray.length == 3) {
                     newReference.setMonument(monument);
 
                     monument.getReferences().add(newReference);
+                    break;
+                case 25: // Image filename
+                    value = formatJpgImageFileName(value);
```
benvogler: Why are we assuming everything is a jpg?

src/main/java/com/monumental/services/AwsS3Service.java
njd9419: Alright, I'll remove the `storeObject()` function.

njd9419: Once my other PR gets merged that moved all of the classes in `/util` and added the `StringHelper` class, I'll move `getObjectUrl` into that class.

src/main/java/util/csvparsing/CsvMonumentConverter.java
njd9419: All of the images I looked at from the initial set were jpgs. I wasn't sure what the valid assumption to make was, so I just chose jpg.

benvogler: While this works now it won't work in the future as I don't think we should restrict users to uploading jpgs. Imagine how annoying that would be

njd9419: Oh no, definitely not. I did it because some of the image filenames in the Initial Dataset spreadsheet didn't have file extensions on them, so the Object URL that would have gotten generated wouldn't have actually existed and the Monument would have no image.

src/main/java/com/monumental/services/AwsS3Service.java
njd9419: Done.

PR #52, opened 2019-11-04
Optimize search query to prevent in memory pagination

benvogler:
There's a pretty annoying "quirk" with JPA 2 where you can't do joins when you're using a limit clause, which causes JPA to then do the limit in memory.

In other words, because we were getting tags alongside the monuments, JPA was querying the entire monument table and then doing a limit in memory, which is terrible and defeats the entire purpose of pagination.

To fix this, we now do the first monument query with the limit but don't get any of the tags. Then, we do another monument query using an IN clause for the ids from the previous query and join the tag table, giving us the same monuments with their tags, which we can then set on the first list of monuments.

End result, we get limited monuments (done through SQL instead of in memory in Java), and we get their related tags.

This also fixes a bug where the count query was failing

### DIFF
--- a/src/main/java/com/monumental/services/MonumentService.java
+++ b/src/main/java/com/monumental/services/MonumentService.java
@@ -1,13 +1,15 @@
 package com.monumental.services;
 
 import com.monumental.models.Monument;
+import com.monumental.models.Tag;
 import com.vividsolutions.jts.geom.Geometry;
 import org.springframework.stereotype.Service;
 
 import javax.persistence.criteria.*;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
-
-import static com.monumental.util.string.StringHelper.isNullOrEmpty;
+import java.util.Map;
 
 @Service
 public class MonumentService extends ModelService<Monument> {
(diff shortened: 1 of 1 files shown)


src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -157,8 +159,41 @@ public Integer countSearchResults(String searchQuery, String latitude, String lo
...
         query.select(builder.countDistinct(root));
 
-        this.buildSearchQuery(builder, query, root, searchQuery, latitude, longitude, distance);
+        this.buildSearchQuery(builder, query, root, searchQuery, latitude, longitude, distance, false);
 
         return this.getEntityManager().createQuery(query).getSingleResult().intValue();
     }
+
+    /**
+     * Gets the related monuments and sets them on the monument objects, using only one extra SQL query
```
njd9419: Shouldn't this be `Gets the related tags and..."?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -157,8 +159,41 @@ public Integer countSearchResults(String searchQuery, String latitude, String lo
...
+
+        Map<Integer, List<Tag>> tagsMap = new HashMap<>();
+        for (Monument monument : monumentsWithTags) {
+            tagsMap.put(monument.getId(), monument.getTags());
+        }
+
+        for (Monument monument : monuments) {
+            monument.setTags(tagsMap.get(monument.getId()));
+        }
+    }
```
njd9419: What's the point of using `tagsMap` and `setTags`?

Won't `monumentsWithTags` return all of the `Monument`s with their `Tag`s already?

benvogler: Yes but they're in the wrong order and not paginated.

PR #54, opened 2019-11-07
Add support for photo sphere images

benvogler:
NOTE: This is branched off of the other PR that's up, so once that PR is merged this one's base branch should be merged until master. Until that happens it shouldn't be merged but can be reviewed.

This PR adds support for 360 degree (Photo Sphere) images, because I think they're pretty nifty.

Preview here: https://giant.gfycat.com/ExcellentWellwornHornshark.webm

Two major todos on this:
1. We need to prevent these from ever being marked as primary, as they will not render correctly as a thumbnail
2. The way this renders them will NOT support 360 images that are uploaded directly to us, it will only support those hosted on google maps currently. Down the road, we could look at [React 360](https://facebook.github.io/react-360/) as a way to render our own 360 images


src/main/java/com/monumental/ExceptionHandler.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+
+/**
+ * This class is responsible for intercepting all exceptions and deciding what should happen with the HTTP response
+ * It also is able to log these exceptions, which is really helpful for debugging since Spring doesn't always print
+ * stacktraces for most exceptions
+ */
+@ControllerAdvice
+public class ExceptionHandler {
```
benvogler: I got fed up with not seeing exception traces again and brought this back. We need to work on the JSON output that it sends to the client because we're not getting useful error messages on the front end currently

njd9419:
The other PR has been merged.

njd9419:
Looks good. Feel free to merge it once you rebase.

PR #55, opened 2019-11-09
Add tag searching to the back-end

benvogler:
Back-end only implementation of filtering monuments by matching tag names


src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -104,9 +134,8 @@ private Predicate buildDWithinQuery(CriteriaBuilder builder, CriteriaQuery query
      */
```
njd9419: Can you add an `@param tags` to this comment?

PR #56, opened 2019-11-10
Related and Nearby Monuments

njd9419:
This PR adds Related and Nearby Monuments to the View M&M Page. This was my first full dive into Redux, so be nice.

I also fixed an issue in the CsvMonumentConverter where we were saving Images for Monuments that had no Image in the spreadsheet.

**Screenshots:**
image

image


benvogler:
Maybe instead of showing "No related Monuments or Memorials" just hide the whole section including the header

client/src/util/api-util.js
```diff
@@ -0,0 +1,8 @@
+export async function get(url, queryString) {
+    let error = null;
+    let res = await fetch(url + queryString)
+        .then(res => res.json())
+        .catch(err => error = err);
+    if (error || res.error) throw(error || res.error);
+    else return res;
+}
```
njd9419: I put this in a separate file so we can re-use it.

njd9419:
@benvogler It will no longer show the Related/Nearby Monuments or Memorials sections if there are none.

client/src/components/Monument/RelatedMonuments/RelatedMonuments.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import "./RelatedMonuments.scss";
+import Spinner from "../../Spinner/Spinner";
+import SearchResult from "../../Search/SearchResult/SearchResult";
+import Collapse from "react-bootstrap/Collapse";
+
+/**
+ * Presentational Component for Related Monuments
+ */
+export default class RelatedMonuments extends React.Component {
```
benvogler: I think you could combine RelatedMonuments and NearbyMonuments, and just pass the title text in as a prop

benvogler: Unless there's a difference between them that I missed

njd9419: Yeah, I can do that.

njd9419: Done.

client/src/actions/monument.js
```diff
@@ -34,5 +80,37 @@ export default function fetchMonument(id) {
...
+            dispatch(fetchNearbyMonumentsError(error));
+        }
+
+        const tags = res.tags.map(tag => tag.name);
+        if (tags.length > 0) {
+            queryOptions = {
+                tags: tags,
+                limit: 6
+            };
+            queryString = QueryString.stringify(queryOptions);
```
benvogler: You need to do `QueryString.stringify(queryOptions, {arrayFormat: 'comma'})` because the default array format doesn't work with spring

njd9419: Done.

PR #57, opened 2019-11-11
Tag model fixes, tag/material filtering, materials using tag model, distance filter

benvogler:
There are lots of changes involved here, which I'll comment on in the change list


src/main/java/com/monumental/models/Model.java
```diff
@@ -1,15 +1,11 @@
...
+import org.springframework.data.jpa.domain.AbstractAuditable;
 
 import javax.persistence.*;
-import javax.validation.constraints.NotNull;
-import javax.validation.constraints.Null;
-import java.util.Date;
 
 @MappedSuperclass
-public abstract class Model {
+public abstract class Model extends AbstractAuditable<User, Integer> {
```
benvogler: Why did no one tell me this existed in 343 😢

benvogler: Also in order to use this I had to make an empty User model

src/main/java/com/monumental/models/Reference.java
```diff
@@ -12,9 +12,6 @@
  */
 
 @Entity
-@Table(name = "reference", uniqueConstraints = {
-        @UniqueConstraint(columnNames = "id")
```
benvogler: Don't need this anymore due to `AbstractAuditable`

src/main/java/com/monumental/models/Tag.java
```diff
@@ -15,18 +15,14 @@
...
-})
 public class Tag extends Model implements Serializable {
 
     @Column(name = "name")
     @NotNull(groups = {New.class, Existing.class}, message = "Name can not be null")
     private String name;
 
     @JsonInclude(JsonInclude.Include.NON_NULL)
-    @ManyToMany(cascade = { CascadeType.ALL })
+    @ManyToMany(cascade = { CascadeType.DETACH })
```
benvogler: Don't ask me what this does or my head will explode.

src/main/java/com/monumental/services/TagRepository.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Repository
+@Transactional
+public interface TagRepository extends CrudRepository<Tag, Integer> {
+
+    /**
+     * This is automatically implemented by JPA - it will get all tags by matching name and isMaterial
+     */
+    @Transactional
+    public List<Tag> getAllByNameAndIsMaterial(String name, Boolean isMaterial);
```
benvogler: poggers

src/main/java/com/monumental/services/TagService.java
```diff
@@ -29,10 +37,73 @@ public TagService(SessionFactoryService sessionFactoryService) {
...
+                builder.gt(wordSimilarity, 0.25),
+                builder.equal(root.get("isMaterial"), isMaterial)
+            )
+        );
+        query.orderBy(builder.desc(builder.sum(wordSimilarity, similarity)));
+
+        return this.getWithCriteriaQuery(query, 10);
+    }
+
+    @Transactional
```
benvogler: `@Transactional` is necessary to be able to initialize lazy loaded collections

src/main/java/com/monumental/services/TagService.java
```diff
@@ -29,10 +37,73 @@ public TagService(SessionFactoryService sessionFactoryService) {
...
+            List<Tag> duplicates = this.tagRepository.getAllByNameAndIsMaterial(name, isMaterial);
+            Tag tag;
+            if (duplicates != null && duplicates.size() > 0) {
+                tag = duplicates.get(0);
+            } else {
+                tag = new Tag();
+                tag.setName(name);
+                tag.setIsMaterial(isMaterial);
+            }
+            tag.getMonuments().addAll(monuments);
```
benvogler: This would error if the method wasn't `@Transactional`

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
```diff
@@ -19,16 +27,23 @@
  * Then does validation on each CsvMonumentConverterResult's Monument before inserting it into the database
  * Finally, closes the CsvFileReader
  */
+@Configuration
+@SpringBootApplication
```
benvogler: By making this an application we get the beautiful spring DI stuff

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
```diff
@@ -19,16 +27,23 @@
...
+@ComponentScan("com.monumental")
 public class InitialDataLoadApplication {
+
     public static void main(String[] args) {
-        String pathToDatasetCsv = "C:\\Users\\nickb\\Documents\\Initial Dataset.csv";
+        SpringApplication app = new SpringApplication(InitialDataLoadApplication.class);
+        // Don't start a web server - allows this application to run while the main Application is running
+        app.setWebApplicationType(WebApplicationType.NONE);
```
benvogler: And this prevents the actual web server from starting on port 8080 so you don't get an error if the server is also running (important for the VM to be able to run it)

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
```diff
@@ -19,16 +27,23 @@
...
+        ConfigurableApplicationContext context = app.run(args);
+        String pathToDatasetCsv = "C:\\Users\\Ben\\Downloads\\Initial.csv";
         String csvRow;
         int rowCount = 0;
         ArrayList<CsvMonumentConverterResult> results = new ArrayList<>();
 
-        SessionFactoryService sessionFactoryService = new SessionFactoryService();
-        MonumentService monumentService = new MonumentService(sessionFactoryService);
-        TagService tagService = new TagService(sessionFactoryService);
+        MonumentService monumentService = context.getBean(MonumentService.class);
```
benvogler: Since we can't use autowire here, this is a way to get the services without using dependency injection

client/src/actions/map.js
```diff
@@ -9,7 +11,7 @@ function fetchMonumentsPending() {
 function fetchMonumentsSuccess(monuments) {
     return {
         type: FETCH_MAP_MONUMENTS_SUCCESS,
-        payload: monuments
+        payload: {monuments}
```
njd9419: Any reason this payload is in curly braces and the one for `searchMonumentsSuccess` isn't?

client/src/actions/search.js
```diff
@@ -31,8 +31,9 @@ export default function searchMonuments(options = {}) {
         const queryString = QueryString.stringify(options);
         dispatch(searchMonumentsPending());
         try {
-            const count = await get('/api/search/count/?', queryString);
-            const monuments = await get('/api/search/?', queryString);
+            const count = await get(`/api/search/monuments/count/?${queryString}`);
```
njd9419: How much faster is the count query? Should we be doing this everywhere we make searching API calls or do you think it's not worth it?

client/src/actions/tagsFilter.js
```diff
@@ -0,0 +1,115 @@
+import {
+    FILTER_TAGS_SEARCH_PENDING, FILTER_TAGS_SEARCH_ERROR, FILTER_TAGS_SEARCH_SUCCESS,
+    FILTER_TAGS_LOAD_ERROR, FILTER_TAGS_LOAD_SUCCESS, FILTER_TAGS_LOAD_PENDING,
+    FILTER_MATERIALS_SEARCH_PENDING, FILTER_MATERIALS_SEARCH_SUCCESS, FILTER_MATERIALS_LOAD_PENDING,
+    FILTER_MATERIALS_LOAD_SUCCESS, FILTER_MATERIALS_LOAD_ERROR, FILTER_MATERIALS_SEARCH_ERROR
+} from '../constants';
```
njd9419: Is there any nicer way to do our imports in our actions?

client/src/components/Search/Filters/TagsFilter/TagsFilter.js
```diff
@@ -0,0 +1,140 @@
...
+        // If the tag is being selected, add it to the selected tags and sort them alphabetically
+        if (value) {
+            tag.selected = true;
+            selectedTags.push(tag);
+            selectedTags = selectedTags.sort();
+        // If the tag is being deselected, remove it from the selected tags
+        } else {
+            const index = selectedTags.findIndex(t => t.name === tag.name);
+            if (index >= 0) {
+                [ tag ] = selectedTags.splice(index, 1);
```
njd9419: What does the `[ ]` around `tag` do here?

client/src/components/Search/Search.js
```diff
@@ -11,11 +11,36 @@ import SearchResults from './SearchResults/SearchResults';
...
+        const { onPageChange } = this.props;
+        onPageChange(page);
+    }
+
+    /**
+     * Whenever the monuments list changes (in length or at least one id has changed), scroll the search results
+     * up to the top, so that we see the first of the new results instead of wherever we might have been
+     * scrolled to before
+     */
+    componentDidUpdate(prevProps, prevState, snapshot) {
```
njd9419: What is the `snapshot` param for?

client/src/components/Tags/Tag/Tag.scss
```diff
@@ -0,0 +1,32 @@
+@import "../../../theme";
+
+.tag {
+    background-color: #e8e8e8;
+    color: rgba(0, 0, 0, 0.8);
+    padding: .5em .75em;
+    border-radius: .28571429rem;
```
njd9419: zoinks

client/src/pages/MonumentPage/MonumentPage.js
```diff
@@ -12,6 +12,7 @@ import fetchMonument from '../../actions/monument';
 class MonumentPage extends React.Component {
 
     static mapStateToProps(state) {
+        console.log('state', state);
```
njd9419: Debug print

client/src/pages/SearchPage/SearchPage.js
```diff
@@ -42,51 +46,43 @@ class SearchPage extends React.Component {
...
-                if (value !== this.state[param]) this.setState(state);
-            } catch (err) {}
-        }
-        return parseInt(value);
+    getQueryParams() {
+        const params = {
+            ...QueryString.parse(this.props.history.location.search, {arrayFormat: 'comma'}),
+            ...this.state
+        };
+        if (params.tags && typeof params.tags === 'string') params.tags = [params.tags];
```
njd9419: What is this line for?

client/src/utils/search.js
```diff
@@ -0,0 +1,25 @@
+import * as QueryString from 'query-string';
+
+/**
+ * Global function for performing monument searches
+ * The page will be redirected to the search results page with the proper query params
+ * @param changedParams The query params to override the current (if existent) query params with
+ * @param history       The react-router history prop must be passed in here so that we can use `.push`
+ */
+export default function search(changedParams, history) {
```
njd9419: Could we rename this to be `searchMonuments()`?

src/main/java/com/monumental/models/Contribution.java
```diff
@@ -13,9 +13,6 @@
  */
 
 @Entity
-@Table(name = "contribution", uniqueConstraints = {
-        @UniqueConstraint(columnNames = "id")
-})
```
njd9419: Are these annotations no longer needed? Why?

src/main/java/com/monumental/models/Model.java
njd9419: That's dumb

src/main/java/com/monumental/models/Reference.java
njd9419: Ah, didn't see this before.

src/main/java/com/monumental/models/Tag.java
njd9419: From my first Google, it seems there's not much explanation.

src/main/java/com/monumental/services/MonumentRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+import javax.transaction.Transactional;
+
+@Repository
+@Transactional
+public interface MonumentRepository extends CrudRepository<Monument, Integer> {
+
+}
```
njd9419: Is this being used anywhere right now? Or will that come in your next PR?

src/main/java/com/monumental/services/TagRepository.java
njd9419: freakin pog

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
```diff
@@ -19,16 +27,23 @@
  * Then does validation on each CsvMonumentConverterResult's Monument before inserting it into the database
  * Finally, closes the CsvFileReader
  */
+@Configuration
+@SpringBootApplication
+@ComponentScan("com.monumental")
```
njd9419: What does this do?

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
njd9419: I thought the point of making it a `@SpringBootApplication` meant we could Autowire?

src/test/java/com/monumental/models/unittest/MonumentUnitTests.java
```diff
@@ -160,29 +161,16 @@ public void testMonument_validate_New_NullTitle_NullMaterial() {
...
-        assertFalse(result.isValid());
-
-        List<String> violationMessages = result.getViolationMessages();
-
         assertEquals(1, violationMessages.size());
         assertTrue(violationMessages.contains("Title can not be null"));
     }
 
+    /**
+     * TODO: This no longer works due to switching Model to use AbstractAuditable
```
njd9419: Are we planning on fixing these or just removing them?

src/test/java/com/monumental/models/unittest/MonumentUnitTests.java
```diff
@@ -192,31 +180,31 @@ public void testMonument_validate_New_NotNullTitle_NullMaterial() {
...
-        assertTrue(violationMessages.contains("Material can not be null"));
-        assertTrue(violationMessages.contains("ID can not be specified on insert"));
-    }
+     assertEquals(1, violationMessages.size());
+     assertTrue(violationMessages.contains("ID can not be null on update"));
+     }
+     */
 
+    /**
+     * TODO: Replace this with repositories instead of services
```
njd9419: What services are we using here...?

njd9419:
First pass completed. Lots of questions but the code looks good.

client/src/actions/tagsFilter.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function searchError(isMaterial, error) {
+    return {
+        type: isMaterial ? FILTER_MATERIALS_SEARCH_ERROR : FILTER_TAGS_SEARCH_ERROR,
+        error: error
+    };
+}
+
+export function searchTags(queryString) {
```
bxsmith007: Just curious, but can't you specify what functions you want to export at the end of the file? If so, we could just list the exports at the bottom instead of in the front of the functions. If not, ignore this.

src/main/java/com/monumental/models/Model.java
benvogler: Well, one day it won't be

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
benvogler: We can't use autowire because we're in a static context. You can only use autowire when your class itself is being autowired (`@Service`, `@Component`, etc.) and is in a non-static context (because that's how Spring does the dependency injection)

client/src/actions/map.js
benvogler: I will double check why they're different, but basically I was running into issues where for example `state.searchPage` would be the search results array instead of an object with `monuments: []` so this fixed that. Need to double check if there is a discrepancy or an actual reason why they're different

client/src/actions/search.js
benvogler: We only do the count query because we actually display it on the page. I would not do the query for any other reason (like checking if there's 0, because it's faster to just do the actual search)

client/src/actions/tagsFilter.js
benvogler: we could group the exports in the constants file like:

```
const FILTER_MATERIALS_LOAD = {
    PENDING: ...,
    SUCCESS: ...,
    ERROR
}
export FILTER_MATERIALS_LOAD
``

then you just need to import FILTER_MATERIALS_LOAD and use FILTER_MATERIALS_LOAD.PENDING etc.

client/src/components/Search/Filters/TagsFilter/TagsFilter.js
benvogler: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Array_destructuring

client/src/components/Search/Search.js
benvogler: Autfilled by IntelliJ using the function signature. Here's the documentation on it but it's something we will probably never use https://reactjs.org/docs/react-component.html#componentdidupdate

client/src/utils/search.js
benvogler: I tried but there's already an action named searchMonuments causing some conflicts and I'm too lazy to think of another name

src/main/java/com/monumental/services/MonumentRepository.java
benvogler: Not being used. We will definitely have one for all the models eventually

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
benvogler: component scan tells it to discover autowireable classes in that package. By default it scans the package that the application is in which would mean this one wouldn't see any of the services etc.

src/test/java/com/monumental/models/unittest/MonumentUnitTests.java
benvogler: Temporary, I think I can add this validation back into Model

src/test/java/com/monumental/models/unittest/MonumentUnitTests.java
benvogler: I think I copied this from another place lol and now I can't remember what's wrong with this one

client/src/actions/tagsFilter.js
benvogler: Yes, you can export any variable i.e. `export searchTags;` or `export searchTags as search;` etc. But I prefer the inline syntax when possible

PR #58, opened 2019-11-14
JPA 2 Refactor

benvogler:
This refactors the existing Hibernate codebase to use the vendor-agnostic JPA 2 API (through [Spring Data JPA](https://spring.io/projects/spring-data-jpa))

The API offers much easier ways of performing CRUD against our models, as such a large amount of code was deleted from `ModelService` which is no longer necessary

I also removed some unused controllers and services (and one unit test file for the deleted service)

**Breaking Change**: The Trigger framework I built is now broken because it relies on HIbernate's event system. I will make another PR later updating this to use the JPA event system.


benvogler:
Because of the way that managed entities work in JPA, a number of our CRUD integration tests are kind of useless now. See this comment for why `.save` actually does nothing in most contexts https://stackoverflow.com/a/13589352/10044594

In essence, changes to managed entities are always persisted. The persistence happens when the entity is flushed to the database, which is _usually_ done automatically by JPA under normal circumstances. So testing if a field changes after you save it, or querying for the updated entity, is pointless because the initial reference will automatically be saved, and will also be updated

benvogler:
One small correction: save _does_ generate the id. Probably the only important thing I've found. Still figuring out how to handle listening to saves within JPA effectively. I can already tell that the JPA ways are not going to do what I want them to because they run after flush instead of after actually updating the field in code.

I'm thinking to take the extremely simple approach for our google address stuff and just adding some logic to the setters. If we ever have need for more complicated listeners we can re-investigate

benvogler:
As an example of how the managed entities act, here's two completely equivalent test methods:
```java
Monument monument = makeTestMonument();

Monument result = this.monumentRepository.save(monument);

Optional<Monument> optional = this.monumentRepository.findById(result.getId());
assert(optional.isPresent());
Monument storedMonument = optional.get();

storedMonument.setTitle("Monument 2");
storedMonument.setArtist("Artist 2");

this.monumentRepository.save(storedMonument);

optional = this.monumentRepository.findById(result.getId());
assert(optional.isPresent());
result = optional.get();

assertEquals(storedMonument.getTitle(), result.getTitle());
assertEquals(storedMonument.getArtist(), result.getArtist());
```
```java
Monument monument = makeTestMonument();

this.monumentRepository.save(monument);

monument.setTitle("Monument 2");
monument.setArtist("Artist 2");

assertEquals(monument.getTitle(), "Monument 2");
assertEquals(monument.getArtist(), "Artist 2");
```
I think I'm opposed to testing the basic CRUD now, because JPA is in charge of it instead of `ModelService`

njd9419:
I see that you're using `.save()` for both creating and updating, JPA handles that automatically?

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -1,47 +1,55 @@
...
+            this.monumentService.initializeAllLazyLoadedCollections(monument);
         }
-
         return monument;
     }
 
     @GetMapping("/api/monuments")
     public List<Monument> getAllMonuments() {
-        return this.monumentService.getAll();
+        return this.monumentRepository.findAll();
```
njd9419: We don't want to expose the ability to get all of the Monuments with lazy loaded collections?

src/main/java/com/monumental/models/Monument.java
```diff
@@ -78,16 +75,12 @@ public Monument() {
     }
 
     public Monument(String artist, String title, Date date, String city, String state) {
+        this();
```
njd9419: What does this do?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -250,13 +253,15 @@ private void getRelatedRecords(List<Monument> monuments, String fieldName) {
...
         String capitalizedFieldName = fieldName.substring(0, 1).toUpperCase() + fieldName.substring(1);
 
         Map<Integer, List> map = new HashMap<>();
         for (Monument monument : monumentsWithRecords) {
             try {
                 map.put(monument.getId(), (List) Monument.class.getDeclaredMethod("get" + capitalizedFieldName).invoke(monument));
+                System.out.println("put " + monument.getId() + " " + map.get(monument.getId()));
             } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
                 System.err.println("Invalid field name: " + fieldName);
                 System.err.println("Occurred while trying to use getter: get" + capitalizedFieldName);
```
njd9419: Would it be such a bad thing to have these methods inside the new Repository classes instead?

src/main/java/com/monumental/services/TagService.java
```diff
@@ -86,24 +45,24 @@ public TagService(SessionFactoryService sessionFactoryService) {
...
+        } else {
+            tag = new Tag();
+            tag.setName(name);
+            tag.setIsMaterial(isMaterial);
         }
+        tag.getMonuments().addAll(monuments);
+        this.tagRepository.save(tag);
+        return tag;
     }
 }
```
njd9419: Same thing as with `MonumentService`, is there a need for Service classes and Repository classes now?

src/main/java/com/monumental/util/application/InitialDataLoadApplication.java
```diff
@@ -33,16 +33,18 @@
...
 
     public static void main(String[] args) {
+
         SpringApplication app = new SpringApplication(InitialDataLoadApplication.class);
         // Don't start a web server - allows this application to run while the main Application is running
         app.setWebApplicationType(WebApplicationType.NONE);
         ConfigurableApplicationContext context = app.run(args);
-        String pathToDatasetCsv = "/home/monumental/Initial Dataset.csv";
+
+        String pathToDatasetCsv = "/Users/benvogler/Downloads/Initial Dataset.csv";
```
njd9419: 👀

src/test/java/com/monumental/services/integrationtest/MonumentRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,300 @@
...
+        assertEquals(2, results.size());
+
+        for (Monument result : results) {
+            Monument monument = monumentsById.get(result.getId());
+
+            assertEquals(monument.getTitle(), result.getTitle());
+            assertEquals(monument.getArtist(), result.getArtist());
+        }
+    }
+
```
njd9419: Can you fix up the comments and method names so they match the method we're calling? I.e. change `get` to `findById`, etc.

njd9419:
Things look pretty good, just a couple of questions.

At some point soon, we should go back through our tests as I feel like they're in a pretty unstable state right now. We should fix them up as necessary and remove the un-needed ones.

njd9419:
I put a new task in the Product Backlog for going through and cleaning up our tests.

src/main/java/com/monumental/controllers/MonumentController.java
benvogler: We do - the ModelService now has `public fetchAllLazyLoadedCollectitons` exposed which can be called explicitly when necessary

benvogler: This is only necessary for controllers who want all the data to be included in the JSON. Everywhere else, if the method is `@Transactional` (or its class) then it will be able to get whichever fields it needs on the fly just by calling the getter

src/main/java/com/monumental/models/Monument.java
benvogler: Calls the default constructor with no params

src/main/java/com/monumental/services/MonumentService.java
benvogler: You can't since they're interfaces

src/main/java/com/monumental/services/TagService.java
benvogler: Yes, the Services should provide any functionality not supported by the Repository since teh Repository is an interface (the actual implementation of the interface is done by JPA, which we can't add onto unfortunately)

src/test/java/com/monumental/services/integrationtest/MonumentRepositoryIntegrationTests.java
benvogler: See my posts above about the tests, I think this whole file can be scrapped honestly

src/main/java/com/monumental/models/Monument.java
njd9419: duh

src/main/java/com/monumental/services/MonumentService.java
njd9419: Oh yeah forgot about that. Kind of a bummer.

PR #59, opened 2019-11-20
Get related monuments by most matching tags

benvogler:
This fixes the related tags implementation so that monuments with one or more matching tags are returned rather than needing all tags to match, and the monuments with the most matches are displayed first.

This does not affect the tag searching on the search page. I could not find a viable way to do this within that `CriteriaQuery`, but the behavior we have there now is probably sufficient anyway.

### DIFF
--- a/src/main/java/com/monumental/repositories/MonumentRepository.java
+++ b/src/main/java/com/monumental/repositories/MonumentRepository.java
@@ -1,11 +1,13 @@
 package com.monumental.repositories;
 
 import com.monumental.models.Monument;
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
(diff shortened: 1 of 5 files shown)


src/main/java/com/monumental/repositories/MonumentRepository.java
```diff
@@ -20,4 +22,14 @@
...
+    /**
+     * Searches for monuments with matching tags by name, returning those with the most matching tags
+     * @param names - The tag names to match against
+     * @param monumentId - The monument to exclude from search results
+     * @param pageable - Used to give the search a limit
+     * @return Tuples of monuments with their count of matching tags
+     */
+    @Query("select m, count(t.id) as c from Monument m join m.tags t where t.name in :names and m.id <> :id group by m.id order by c desc")
+    List<Tuple> getRelatedMonuments(@Param("names") List<String> names, @Param("id") Integer monumentId, Pageable pageable);
+
```
njd9419: List of Tuples dude

PR #60, opened 2019-11-20
Fix images not loading in related monuments

benvogler:
Do not merge until I get a chance to test this tonight. Can't test right now because I haven't gotten PostGIS working on macos yet.

### DIFF
--- a/client/src/actions/monument.js
+++ b/client/src/actions/monument.js
@@ -102,7 +102,7 @@ export default function fetchMonument(id) {
         if (tags.length > 0) {
             queryOptions = {
                 tags: tags,
-                limit: 6,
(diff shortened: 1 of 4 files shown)


benvogler:
@njd9419 this is ready to go and working locally. Unfortunately I had to change it so that it loads the image collection individually for each monument which is way less efficient but we can revisit that at another time

njd9419:
I added a commit so that the Travis build would kick off again.

njd9419:
For some reason Travis isn't saying that the build completed even though it has.

benvogler:
I turned off the travis requirement for now. Hopefully just a blip on this PR...

PR #61, opened 2019-12-01
Create M+M Page

njd9419:
This PR is a bit larger in scope than originally anticipated, but the extra things I addressed I felt were warranted because I was already touching that area of the code.

This PR adds:
1. A Create M+M Page with form validation and S3 integration for images
2. Moves some of the duplicated logic into the Service classes
3. Removes/Adds/Fixes all of our tests as needed


client/src/actions/tagsSearch.js
```diff
@@ -1,29 +1,29 @@
 import {
```
njd9419: The changes in this file are all renames stemming from making the `TagsSearch` component independent of the `Filters` component.

client/src/components/Search/TagsSearch/TagsSearch.js
```diff
@@ -40,30 +40,59 @@ class TagsFilter extends React.Component {
...
     }
 
     render() {
-        const { searchQuery, selectedTags } = this.state;
-        const { searchResults, variant } = this.props;
+        const { searchQuery, selectedTags, createdTags, createdTagsKey } = this.state;
+        const { searchResults, variant, className, allowTagCreation } = this.props;
         const visibleSearchResults = searchResults.filter(result => {
             return !selectedTags.find(tag => tag.id === result.id);
         });
```
njd9419: This section below adds the ability to create a new Tag/Material from the `TagsSearch` component.

client/src/components/Search/TagsSearch/TagsSearch.js
```diff
@@ -73,19 +102,22 @@ class TagsFilter extends React.Component {
...
+                    {createNewTagDisplay}
                 </div>
             </div>
         );
     }
 
     async handleSelectTag(value, tag) {
         let { selectedTags } = this.state;
+        const { createdTags } = this.state;
+        const { variant, onChange } = this.props;
```
njd9419: This component also takes a few more props now so selecting/deselecting a Tag/Material can be managed by the parent (See `Filters` for an example).

src/main/java/com/monumental/models/api/CreateMonumentRequest.java
```diff
@@ -0,0 +1,170 @@
+package com.monumental.models.api;
```
njd9419: I had to use this class to avoid the JSON deserializer attempting to create Model classes such as Tags, References, etc without all of the values being present.

client/src/App.js
```diff
@@ -11,6 +11,7 @@ import SearchPage from './pages/SearchPage/SearchPage';
 import ErrorHandler from './containers/ErrorHandler/ErrorHandler';
 import Toaster from './containers/Toaster/Toaster';
 import MapPage from './pages/MapPage/MapPage';
+import CreatePage from "./pages/CreatePage/CreatePage";
```
benvogler: We should probably call this `CreateMonumentPage`

client/src/actions/tagsSearch.js
benvogler: Not really an issue for right now, but this means the `TagsSearch` will not work if there's more than one on the same page (of the same type)

client/src/components/CreateForm/CreateForm.scss
```diff
@@ -0,0 +1,175 @@
...
+      box-shadow: 0 0 0 0.2rem rgba(66, 184, 131, 0.25);
+    }
+
+    &:last-child {
+      margin-right: 0;
+    }
+
+    &.is-invalid {
+      border-color: #da6060;
+      background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='%23da6060' viewBox='-2 -2 7 7'%3e%3cpath stroke='%23da6060' d='M0 0l3 3m0-3L0 3'/%3e%3ccircle r='.5'/%3e%3ccircle cx='3' r='.5'/%3e%3ccircle cy='3' r='.5'/%3e%3ccircle cx='3' cy='3' r='.5'/%3e%3c/svg%3E");
```
benvogler: Could you give a comment saying what this is an icon of or where you got it. Also by the way we have material icons loaded on the site, so if it has something that covers this you should use it https://material.io/resources/icons/?style=baseline

client/src/components/CreateForm/NoImageModal/NoImageModal.js
```diff
@@ -0,0 +1,45 @@
...
+            <Modal
+                show={showing}
+                onHide={onClose}
+            >
+                <Modal.Header closeButton>
+                    <Modal.Title>No Images Uploaded</Modal.Title>
+                </Modal.Header>
+                <hr/>
+                <Modal.Body>
+                    <p>At Monuments and Memorials, we're trying our best to provide the most informative data on all of our records.</p>
```
benvogler: This makes it sound like "Monuments an Memorials" is an actual organization or something. I would replace this with simply "We try our best to..."

client/src/utils/api-util.js
```diff
@@ -1,8 +1,78 @@
...
     if (error || res.error) throw(error || res.error);
     else return res;
+}
+
+/**
+ * Send a POST request to the specified url with the specified data
+ * @param url - URL to send the POST to
+ * @param data - JSON data to send to the specified URL
+ */
+export async function post(url, data) {
```
benvogler: doesn't have to be done now, but we should probably abstract this into a private function that accepts the method as a param, sot hat we can easily do PUT and DELETE

src/main/java/com/monumental/models/Model.java
```diff
@@ -22,6 +24,8 @@
     }
 
     @AttributeOverride(name = "id", column = @Column(name="id"))
+    @Null(groups = New.class, message = "ID can not be specified on insert")
+    @NotNull(groups = Existing.class, message = "ID can not be null on update")
```
benvogler: thank

src/main/java/com/monumental/models/Monument.java
```diff
@@ -74,13 +74,18 @@ public Monument() {
         this.contributions = new ArrayList<>();
     }
 
-    public Monument(String artist, String title, Date date, String city, String state) {
+    public Monument(String artist, String title, Date date, String city, String state, String address,
```
benvogler: we should probably just ditch this whole constructor honestly and just use the setters

njd9419:
Don't merge yet please. Going to tackle some of the things Ben mentioned before merging.

client/src/components/CreateForm/CreateForm.scss
njd9419: This is actually stolen from the styling that `react-bootstrap` uses for invalid Form Controls. I had to modify the CSS a bit to get it to work with `TagsSearch` since it's not a true Form Control, hence this.

client/src/components/CreateForm/NoImageModal/NoImageModal.js
njd9419: Coming in the next commit.

client/src/actions/tagsSearch.js
njd9419: Why do you say that?

client/src/utils/api-util.js
njd9419: Sounds like a good idea. Don't feel like doing it now.

src/main/java/com/monumental/models/Monument.java
njd9419: Coming in the next commit.

client/src/App.js
njd9419: Coming in the next commit.

PR #65, opened 2019-12-03
Implement distance and date sorting in search results

benvogler:
This PR adds distance and date (newest first or oldest first) sorting to the search page, and also makes the existing dropdown functional.

I'm not gonna lie, I don't really know what the default sort is if you do relevance without a text search. I think it's probably based on when the row was inserted into the database, probably oldest first
image

Distance sort
image

Oldest (Canton Viaduct is from 1835)
image

Newest (Boston Irish Famine Memorial is from 1998)
image

### DIFF
--- a/client/src/pages/SearchPage/SearchPage.js
+++ b/client/src/pages/SearchPage/SearchPage.js
@@ -52,7 +52,7 @@ class SearchPage extends React.Component {
                 <Spinner show={pending}/>
                 <Search monuments={monuments} {...this.getQueryParams()} count={count}
                         onLimitChange={this.handleLimitChange.bind(this)} onPageChange={this.handlePageChange.bind(this)}
-                        onFilterChange={this.handleFilterChange.bind(this)}/>
+                        onFilterChange={this.handleFilterChange.bind(this)} onSortChange={this.handleSortChange.bind(this)}/>
             </div>
         );
     }
@@ -77,6 +77,12 @@ class SearchPage extends React.Component {
     handleFilterChange(filters) {
         this.search({
             d: filters.distance
+        });
+    }
+
+    handleSortChange(sort) {
+        this.search({
+            sort
         });
     }
 
--- a/src/main/java/com/monumental/services/MonumentService.java
+++ b/src/main/java/com/monumental/services/MonumentService.java
@@ -41,6 +41,13 @@ public class MonumentService extends ModelService<Monument> {
     public static final int feetSrid = 2877;
 
     /**
+     * This enum is used when choosing how to sort search results
+     */
+    public enum SortType {
+        RELEVANCE, DISTANCE, NEWEST, OLDEST, NONE;
+    }
+
+    /**
      * Builds a similarity query on the Monument's title, artist and description fields, and adds them to your CriteriaQuery
      * @param builder           Your CriteriaBuilder
      * @param query             Your CriteriaQuery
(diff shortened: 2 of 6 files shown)


src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -75,28 +82,45 @@ private Predicate buildSimilarityQuery(CriteriaBuilder builder, CriteriaQuery qu
...
+            ),
+            builder.function("ST_Transform", Geometry.class,
+                builder.function("ST_GeometryFromText", Geometry.class,
+                    builder.literal(comparisonPointAsString),
+                    builder.literal(coordinateSrid)
+                ),
+                builder.literal(feetSrid)
+            ),
+            builder.literal(feet)
+        };
```
njd9419: Can you add some comments to this section so it's easier to see what `expressions[0]` is in the inline array declaration?

njd9419:
First pass completed: Things look good. Just going to take one more look through.

client/src/pages/SearchPage/SearchPage.js
```diff
@@ -80,6 +80,12 @@ class SearchPage extends React.Component {
         });
     }
 
+    handleSortChange(sort) {
+        this.search({
+            sort
+        });
```
njd9419: The way you wrote `SearchPage.search()` is genius.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -154,19 +178,38 @@ private Predicate buildTagsQuery(CriteriaBuilder builder, CriteriaQuery query, R
...
+
+        switch (sortType) {
+            case NEWEST:
+                this.sortByDate(builder, query, root, true);
+                break;
+            case OLDEST:
+                this.sortByDate(builder, query, root, false);
+                break;
+            case DISTANCE:
+                sortByDistance = true;
```
njd9419: Missing `break;` here?

njd9419: Nevermind, you probably intentionally did it that way so searches with distance and text can be ordered by both.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -154,19 +178,38 @@ private Predicate buildTagsQuery(CriteriaBuilder builder, CriteriaQuery query, R
...
+        }
+
         if (!isNullOrEmpty(searchQuery)) {
-            predicates.add(this.buildSimilarityQuery(builder, query, root, searchQuery, 0.1, orderByResults));
+            predicates.add(this.buildSimilarityQuery(builder, query, root, searchQuery, 0.1, sortByRelevance));
         }
 
         if (latitude != null && longitude != null && distance != null) {
-            predicates.add(this.buildDWithinQuery(builder, root, latitude, longitude, distance));
+            predicates.add(this.buildDWithinQuery(builder, query, root, latitude, longitude, distance, sortByDistance));
```
njd9419: I'm a bit confused about what happens if a text AND location search happens and the `SortType` is `DISTANCE`. I **think** we'll first order the results by relevance, then distance? Is that right? And if so, is that what behavior we want?

njd9419:
Second pass completed: One question added and a request for comments.

benvogler:
@njd9419 ignore the request for review, I missed your second pass. Working on those now

src/main/java/com/monumental/services/MonumentService.java
benvogler: This shouldn't ever have 2 sorts happening. That *may* have been possible before because I was missing breaks in the switch statement, but now you should only ever have `sortByDistance` OR `sortByRelevance` not both.

src/main/java/com/monumental/services/MonumentService.java
benvogler: Fixed. Also let's not talk about the genius that is
```java
default:
    break;
```

client/src/pages/SearchPage/SearchPage.js
benvogler: The way I write everything is genius. Especially bugs. Genius bugs. We have the best bugs. The biggest bugs.

src/main/java/com/monumental/services/MonumentService.java
benvogler: I changed them to variables because what even in the hell was I doing before, I don't know

benvogler:
@njd9419 Ok, ready for review again fo real

src/main/java/com/monumental/services/MonumentService.java
njd9419: Alright yes, it was because you were missing `break`s. I thought at first that it might have been intentional, which is why I asked.

PR #67, opened 2019-12-12
Bulk Monument Create

njd9419:
This PR adds the ability to upload a CSV file for bulk creation of Monuments.

I leveraged the functionality we already had written for the `InitialDataLoadApplication` and placed it inside `MonumentService` so we could re-use it in both places.

I also added unit tests in places we were lacking them.


client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -0,0 +1,226 @@
...
+        return (
+            <div className='bulk-create-form-container'>
+                <div className='h5'>
+                    Bulk Create Monuments and Memorials
+                </div>
+
+                <Form onSubmit={(event) => this.handleSubmit(event)}>
+                    <Form.Group className='file-upload-form-group'>
+                        <Form.Label>Upload a CSV File:</Form.Label>
+                        <label htmlFor='file-upload-input' className='file-upload-input-label'>
```
njd9419: I overlaid a `<label>` on this HTML file input because you can't style the input.

src/main/java/com/monumental/models/Monument.java
```diff
@@ -246,11 +243,13 @@ public String toString() {
     private String generateDescription() {
```
njd9419: This method is a bit of a mess right now. There's a ticket in our backlog to clean it up.

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -59,76 +74,71 @@ public static CsvMonumentConverterResult convertCsvRow(String csvRow) {
...
+                case 12: // Reference
                     Reference newReference = new Reference();
                     newReference.setUrl(value);
                     newReference.setMonument(monument);
 
                     monument.getReferences().add(newReference);
                     break;
-                case 25: // Image filename
+                    // TODO: Support images in Bulk Monument Creation
+                /*case 25: // Image filename
```
njd9419: Bulk Create with Images is coming soon.... 👀

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverterResult.java
```diff
@@ -77,4 +86,220 @@ public void addTag(Tag tag) {
 
         this.tags.add(tag);
     }
+
+    /**
+     * Validates this CsvMonumentConverterResult
+     * If any of the validations fail, this CsvMonumentConverterResult is considered invalid
+     * @return ValidationResult - ValidationResult object representing the result of the validation
+     */
+    public ValidationResult validate() {
```
njd9419: I decided to move the validation out of `Monument` and place it here instead. It makes much more sense to me to do it this way before we ever create a `Monument`.

njd9419: Plus, this is a friendly way to get the validation results back to the front-end and make them easy to use.

PR #70, opened 2019-12-23
Tag Directory Page

njd9419:
This PR adds a Tag Directory Page to the site. It's pretty minimal right now, but will probably look a bit nicer once we get some feedback from Dr. Decker.

Perhaps adding filtering to this page may be a good idea in the future.

Screenshot:
image


client/src/actions/tagDirectory.js
```diff
@@ -0,0 +1,35 @@
+import { FETCH_TAGS_PENDING, FETCH_TAGS_SUCCESS, FETCH_TAGS_ERROR } from '../constants';
```
benvogler: Is it a good idea to reuse the constants? doesn't this mean that the reducer will affect the other places we use these constants?

client/src/components/TagColumn/TagColumn.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import './TagColumn.scss';
+import Tags from '../Tags/Tags';
+
+/**
+ * Presentational component for a column of Tags
+ */
+export default class TagColumn extends React.Component {
```
benvogler: Why is this component necessary?

client/src/pages/TagDirectoryPage/TagDirectoryPage.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    static mapStateToProps(state) {
+        return state.tagDirectoryPage;
+    }
+
+    render() {
+        const { tags, fetchTagsPending } = this.props;
+
+        const allTags = [];
```
benvogler:
```suggestion
        const allTags = tags.filter(tag => !tag.isMaterial);
```

client/src/pages/TagDirectoryPage/TagDirectoryPage.js
```diff
@@ -0,0 +1,67 @@
...
+        const allMaterials = [];
```
benvogler:
```suggestion
        const allMaterials = tags.filter(tag => tag.isMaterial);
```

client/src/actions/tagDirectory.js
njd9419: They are not being re-used. I added them to `constants.js` down below.

njd9419: I can make them a little more specific to the Tag Directory Page though if you think that's needed.

client/src/components/TagColumn/TagColumn.js
njd9419: When I first started I thought it'd be more, but it's not. I'll remove it.

client/src/actions/tagDirectory.js
benvogler: Ah my bad. Maybe they should be more specific but it's not super important

PR #71, opened 2019-12-25
Bulk M+M Creation with Images

njd9419:
This PR adds support for image uploads during bulk M+M creation.

Some of the implementation got a little wacky with having to send the file to the back-end and do conversions between `ZipEntry`s and `MultipartFile`s and `File`s etc.


njd9419:
FYI: Some of the error handling in this PR is not 100% correct and changes in my next PR after this one.

src/main/resources/application.properties
```diff
@@ -19,4 +19,7 @@ spring.jpa.properties.hibernate.connection.driver_class=org.postgresql.Driver
 spring.jpa.properties.hibernate.jdbc.lob.non_contextual_creation=true
 spring.jpa.show-sql = true
 spring.jpa.hibernate.ddl-auto = update
-spring.jpa.properties.hibernate.default_schema=public
\ No newline at end of file
+spring.jpa.properties.hibernate.default_schema=public
+spring.servlet.multipart.enabled=true
+spring.servlet.multipart.max-file-size=300MB
+spring.servlet.multipart.max-request-size=300MB
```
njd9419: Is this too large? Not sure what the standard is.

njd9419:
A discussion of the implementation:

There are now 2 ways to bulk create monuments:

**1. CSV File Upload**
- In this scenario, the CSV file contents are read on the front-end and a List of CSV Strings are sent to the server for processing.

**2. .zip File Upload**
- In this scenario, the entire .zip file is sent from the front-end to the server because it is near impossible to process .zip files on the front-end.
- Once the back-end receives the .zip file, it searches it for one CSV file. If it does not find exactly one CSV file, an error occurs.
- While searching for the CSV file, if any of the other files scanned are supported image files (currently only .jpg and .png files), their filenames are added to a list. Any other files are ignored.
- Next, we pull the CSV file contents out into a List of CSV Strings like we do in scenario 1 and begin to process it.
- However, we do some "pre-processing" on the row before it goes to the normal bulk create processing: If the value in the image filename column of the CSV row matches one of the image filenames we collected, we upload that image to S3 and set the value of the column in the row to the full S3 Object URL. If the value does not match, we set it to blank. This is done so we don't have to iterate the entire file contents twice. Now, when the "normal" bulk create processing occurs, it'll just save whatever value is in the image filename column (if it is not blank) to the Image table.

njd9419:
Like I said, I realize the implementation got a little yucky at some points so feel free to ask any questions.

src/main/resources/application.properties
benvogler: Yes this is too large. I would start at 10MB. A normal image size is like <100KB

PR #73, opened 2020-01-09
About Page Implementation

njd9419:
This PR adds the About Page to the site. This page was created based on Dr. Decker's request in the document she sent us about the Home Page and About Page contents (link here: https://docs.google.com/document/d/16jemVa-WVvC0FeSkrKncaayf6dWu3aqZ/edit?dls=true)

**Screenshots:**

image

image

image

image


client/src/components/About/AboutInformation/AboutInformation.js
```diff
@@ -0,0 +1,159 @@
...
+                    About Monuments + Memorials
+                </h1>
+                <p>
+                    <span className='font-italic'>Monuments + Memorials</span> is a crowd-sourced initiative developed at
+                    the Rochester Institute of Technology under the direction of Dr. Juilee Decker. Students in her spring
+                    2019 history course <span className='font-italic'>Monuments & Memory</span> began data collection by
+                    examining 26 states or territories in the US. The following fall, software engineering students in the
+                    Golisano College of Computer and Information Sciences built this web application to house data and to
+                    generate further interest and data collection. To view monuments and memorials, {mapNavLink}. To enter
+                    data, click here.
```
njd9419: This `click here` will have to be replaced with a link when we have a sign in page.

client/src/components/About/AboutInformation/AboutInformation.js
```diff
@@ -0,0 +1,159 @@
...
+                    Our goal is to map every monument and memorial in the United States and its inhabited territories
+                    (American Samoa, Guam, Puerto Rico, the North Mariana Islands, and U.S. Virgin Islands). We invite
+                    anyone to contribute information and an image of monuments and memorials in these areas. If you have
+                    any questions, please contact the project coordinator, Juilee Decker jdgsh@rit.edu.
+                </p>
+                <h3 className='font-italic'>
+                    Contributors:
+                </h3>
+                <p>
+                    We invite contributions from anyone. If you are interested in contributing, please click here. Your
```
njd9419: This `click here` will also have to be replaced with a link to the sign-in page when that exists.

client/src/components/About/AboutInformation/AboutInformation.js
benvogler: Can you add a todo comment for this

src/main/java/com/monumental/models/api/MonumentAboutPageStatistics.java
```diff
@@ -0,0 +1,91 @@
+package com.monumental.models.api;
+
+import com.monumental.models.Monument;
+
+import java.util.HashMap;
+
+/**
+ * Model class that holds statistics about the Monuments in the system for the About Page
+ */
+public class MonumentAboutPageStatistics {
```
benvogler: I don't think this necessarily belongs in the model package if it doesn't extend `Model`, although I'm not sure what the best place for it is

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -647,4 +653,85 @@ private String preProcessImageForCsvRow(String csvRow, List<String> imageFileNam
             return CsvMonumentConverter.setImageFileNameOnCsvRow(csvRow, "");
         }
     }
+
+    /**
+     * Gathers the various statistics related to Monuments for the About Page
+     * @return MonumentAboutPageStatistics - Object containing the various statistics relating to Monuments for the
+     * About Page
+     */
+    public MonumentAboutPageStatistics getMonumentAboutPageStatistics() {
```
benvogler: In a perfect world this is something we would cache and only update every day or so, but that's a lot of effort to implement

src/main/java/com/monumental/util/string/StringHelper.java
```diff
@@ -43,4 +43,197 @@ public static String removeBeginningAndEndingQuotes(String string) {
...
+     * Method to parse a state abbreviation or full name into a common format
+     * @param state - State String to parse into common format
+     * @return String - The name of the State in a common format
+     */
+    public static String parseState(String state) {
+        if (state == null) {
+            return null;
+        }
+
+        switch (state.toLowerCase()) {
```
benvogler: I can't tell if I hate this for having 100 cases or like it for being simple and not requiring a dependency

benvogler:
I have some stylistic changes I want to make but we can do that in a later PR

client/src/components/About/AboutInformation/AboutInformation.js
njd9419: Will do.

client/src/components/About/AboutInformation/AboutInformation.js
njd9419: Will do.

src/main/java/com/monumental/models/api/MonumentAboutPageStatistics.java
njd9419: I could move them into a subfolder of `com.monument.controllers`, something like `helpers` or something. Thoughts?

src/main/java/com/monumental/services/MonumentService.java
njd9419: Would be pretty cool though.

src/main/java/com/monumental/util/string/StringHelper.java
njd9419: I figured it was easier to roll our own 🤷‍♂

src/main/java/com/monumental/models/api/MonumentAboutPageStatistics.java
benvogler: Fine by me

njd9419: Done.

PR #75, opened 2020-01-14
Mobile, Tablet, and Desktop map page

bxsmith007:


client/src/pages/MapPage/MapPage.js
```diff
@@ -3,15 +3,31 @@ import './MapPage.scss';
 import MapResults from '../../components/Search/MapResults/MapResults';
 import { connect } from 'react-redux';
 import fetchMonuments from '../../actions/map';
+import {zoom} from "leaflet/src/control/Control.Zoom";
```
njd9419: This is an un-used import. Please remove.

client/src/pages/MapPage/MapPage.js
```diff
@@ -3,15 +3,31 @@ import './MapPage.scss';
...
 
 class MapPage extends React.Component {
 
     constructor(props) {
         super(props);
+        this.state = {
+            zoomSize: '5'
+        }
```
njd9419: Minor: missing `;`

client/src/pages/MapPage/MapPage.js
```diff
@@ -3,15 +3,31 @@ import './MapPage.scss';
...
 
+    componentWillMount() {
+        let zoomSize;
+        if(window.innerWidth < 800) {
+            zoomSize = '3';
+        } else if(window.innerWidth < 1600) {
+            zoomSize = '4';
+        } else {
+            zoomSize = '5';
+        }
```
njd9419: This is an unnecessary `else` statement.

I'd just refactor this to be:

`componentWillMount() {
   if (window.innerWidth < 800) {
      this.setState({zoomSize: 3});
   }
   else if (window.innerWidth >= 800 && window.innerWidth < 1600) {
      this.setState({zoomSize: 4});
   }
}`

njd9419:
First pass completed.

Won't these changes just change the level at which the map zooms? Will it change the aspect ratio for phones and tablets?

Some screenshots would be helpful.

client/src/pages/MapPage/MapPage.js
bxsmith007: In the state, I don't think it uses a ' ; '

If there were more things in the list, then it would use a ' , ' I believe

bxsmith007:
Yes, it basically just sets the initial zoom distance when first opening up the map. I don't think the page really needs more than that, and it looks pretty good on all platforms. Ill get some pictures asap

client/src/pages/MapPage/MapPage.js
njd9419: I was referring to line 14, after the `}`.

bxsmith007: Oh, sorry XD

bxsmith007:
WidthSmallerThan700

The above image is smaller than 700px wide
Zoom: 3

WidthAbove700_Below1500

The above image is bigger than 700px, but smaller than 1600px
Zoom: 4

bxsmith007:
Wait, I just realized that the zoom in inspect was not at 100% for these. That distorted the image to look appropriate when it wasn't. I am making some adjustments now.

bxsmith007:
WidthBelow700_360px

The above image is 360px wide and like 550px tall
Zoom: 3

WidthJustAbove700

This image is just above 700px wide and 530px tall
Zoom: 4

njd9419:
This looks good to me, but I want to defer to @benvogler.

bxsmith007:
Does this need to be changed anymore? There wasn't any styling that I had on the page, so I don't think I need to change anything to a scss file

njd9419:
I personally don't think so, but liked I said I want @benvogler to approve it since I'm not really familiar with this stuff.

PR #76, opened 2020-01-14
Mobile, Tablet, and Desktop search page

bxsmith007:


client/src/components/Search/Search.js
```diff
@@ -46,28 +81,82 @@ export default class Search extends React.Component {
...
+                                    <SearchResults monuments={monuments} limit={limit} page={page}/>
+                                </div>
+                                <div className="pagination-container">
+                                    <Pagination count={pageCount}
+                                                page={page - 1}
+                                                onPage={page => this.handlePageChange(page + 1)}/>
+                                </div>
+                            </div>
+                        </>
+                    }
```
njd9419: Can't this be done via media breakpoints inside `.scss` files? This is a mess and makes it extremely hard to edit in the future.

client/src/components/Search/Search.scss
```diff
@@ -37,6 +50,47 @@
...
 }
 
+.search-row {
+    width: 100%;
+    flex: 0 0 100%;
+    @include media-breakpoint-up(lg) {
+        max-width: none;
+        width: 100%;
+    }
+    //max-width: calc(100vw - 2rem);
```
njd9419: Remove or uncomment

client/src/components/Search/Search.scss
```diff
@@ -37,6 +50,47 @@
...
+    height: 60%;
+    overflow-y: auto;
+    display: flex;
+    flex-direction: column;
+    //flex-wrap: wrap;
```
njd9419: Remove or uncomment

client/src/components/Search/Search.scss
```diff
@@ -37,6 +50,47 @@
...
+    }
+}
+
+.search-full-page {
+    flex: 0 0 100%;
+    @include media-breakpoint-up(lg) {
+        max-width: none;
+        width: 100%;
+    }
+    //max-width: calc(100vw - 2rem);
```
njd9419: remove or uncomment

client/src/components/Search/Search.scss
```diff
@@ -37,6 +50,47 @@
...
+    height: 100%;
+    overflow-y: auto;
+    display: flex;
+    flex-direction: column;
+    //flex-wrap: wrap;
```
njd9419: remove or uncomment

client/src/components/Search/Search.js
bxsmith007: I don't think so. I could do it for the different class names, but I need to rerender new components or get rid of components at certain widths.

njd9419: But can't you have them `display: block` (or whatever display you want) by default then at different sizes have the `display: none` as you need?

bxsmith007: I never tried it that way, I really only knew how to do this way efficiently. I can try that way and see how it goes.

I still think it needs to be at least 2 things because I am changing where a component is being rendered, not just visibility, but maybe that can also be done in the scss file. I just don't know how to do that efficiently

bxsmith007: I can get it done that way, I just don't know if I can get it done that way by 5. I will try though

njd9419:
Still some conflicts here.

njd9419:
I think this looks good, but @benvogler said he'd like to look at this.

client/src/components/Search/Search.js
```diff
@@ -33,7 +28,9 @@ export default class Search extends React.Component {
                 });
             }
             if (didChange) {
-                document.querySelector('.search-column').scrollTo({top: 0});
+                if(document.getElementById('search-column')) {
+                    document.querySelector('search-column').scrollTo({top: 0});
```
benvogler: This doesn't make any sense, it was selecting a class not an id, and your second selector is missing the `.` before `search-column`

```
let searchColumn = document.querySelector('.search-column');
if (searchColumn) searchColumn.scrollTo({top: 0});
```

client/src/components/Search/Search.js
```diff
@@ -56,31 +53,31 @@ export default class Search extends React.Component {
...
         return (
-                <div className="search-results-page">
-                    <div className="map-column">
-                        <MapResults monuments={monuments} zoom={lat && lon ? 10 : 4} center={lat && lon ? [lat, lon] : null}/>
-                    </div>
-                    <div className="search-column">
-                        <div className="search-header">
-                            <Filters onChange={filters => onFilterChange(filters)}
+            <div className="search-results-page">
+                <div className="map-column d-none d-sm-none d-md-flex">
```
benvogler:
```suggestion
                <div className="map-column d-none d-md-flex">
```

`d-sm-none` was redundant because `d-none` covers it and `d-md-flex` makes everything md and above flex

PR #77, opened 2020-01-16
Refactor header mobile responsiveness

benvogler:
It was difficult to explain in comments the changes I would like to see to the mobile responsiveness PRs, so I thought I would make them to one of the three as an example. The basic steps I took were to:

1. Remove any style attributes and use classes/scss instead because it's much easier to maintain that way
2.  Remove any places style information are being passed into props and instead just control that from the parent component's scss, since the scss is scoped downward and parents can modify the scss of their children
3. Use CSS responsiveness over javascript screen width checks where possible. This is generally preferred. I left in several javascript usages though to avoid duplicated content on the page, because this is very bad for SEO
4. Slight adjustments to some of the look of the search bar, the search/hamburger menu icons, and the hamburger menu

@bxsmith007 I hope this PR can be used to help you update the other 2 PRs, but if there is any confusion just shout and we can get on a call or meet in person to go over this

Desktop

Tablet

Mobile

Mobile search modal

Mobile hamburger menu


client/src/App.scss
```diff
@@ -73,3 +73,22 @@ homepage-map {
     font-weight: bold;
     font-size: 24px;
 }
+
+.mobile-search-modal {
```
benvogler: This stuff needs to be in App.scss because the modal component puts its elements high up in the element tree, above where the Header.scss can target

client/src/components/Header/Header.js
```diff
@@ -34,97 +34,83 @@ export default class Header extends React.Component {
...
+        let headerLinks = this.links.map(link =>
+            <NavLink onClick={e => {
+                e.preventDefault();
+                window.location.replace(link.route);
+            }} to={link.route} exact={link.exact} className="header-link mr-4" activeClassName="active" key={link.name}>{link.name}</NavLink>
+        );
         return (
             <div className="header" id="pageHeader" ref={element => this.divRef = element}>
-                {
-                    this.state.screenWidth > 830 && <>
```
benvogler: Rather than having separate left/tabletLeft/right/tabletRight/etc, there is now one left/center/right div and css breakpoints are used to change styles depending on screen size. I'm still using some javascript screen size checking here when it comes to whether or not something should be rendered at all on that screen, but there's no copy and pasted markup anymore

client/src/components/Header/Header.js
```diff
@@ -34,97 +34,83 @@ export default class Header extends React.Component {
...
-                                        <li><Button size="sm" variant="link-secondary" style={{padding: 0}}>Sign up</Button></li>
-                                    </ul>
-                                </div>
-                            </CheeseburgerMenu>
+                    </div>
+                </div>
 
+                <div className="center">
+                    {/* Desktop and Tablet */}
+                    {window.innerWidth >= 576 && <>
```
benvogler: I also switched the pixel amounts to match the bootstrap breakpoints https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints

client/src/components/Header/Header.js
```diff
@@ -34,97 +34,83 @@ export default class Header extends React.Component {
...
+                        <Button variant="link-secondary" className="search-icon" onClick={() => this.setState({isSearchModalOpen: true})}>
+                            <i className="material-icons">
+                                search
+                            </i>
+                        </Button>
+                    </>}
+                    {/* Tablet and Mobile */}
+                    {window.innerWidth < 992 && <>
+                        <Button variant="link-secondary" className="menu-icon" onClick={() => this.setState({isMenuOpen: true})}>
+                            <i className="material-icons">
```
benvogler: FYI @bxsmith007 we have material icon font loaded on the site so you can just do icons this way - much easier than adding images. Check here for info https://material.io/resources/icons/. If you click an icon, you can see on the left side how to add it via the font

client/src/components/Header/Header.js
```diff
@@ -34,97 +34,83 @@ export default class Header extends React.Component {
...
+                                menu
+                            </i>
+                        </Button>
+                        <CheeseburgerMenu isOpen={this.state.isMenuOpen} closeCallback={() => {this.setState({isMenuOpen: false})}}>
+                            <ul>
+                                <div className="links d-lg-block">
+                                    {headerLinks.map(link =>
+                                        <li>{link}</li>
+                                    )}
```
benvogler: Rather than copy and paste the link code, I did it once in a variable at the start of the render function and then here I just wrap each link in a `<li>` so that it works in the `<ul>`

client/src/components/Header/Header.scss
```diff
@@ -1,37 +1,86 @@
...
+@import "../../../node_modules/bootstrap/scss/variables";
+@import "../../../node_modules/bootstrap/scss/mixins";
+
 .header {
     display: flex;
-    padding: 2rem;
     font-size: 0.875rem;
     position: relative;
+    padding: 1rem;
+    @include media-breakpoint-up(lg) {
```
benvogler: @bxsmith007 this is how you control responsiveness from SCSS using Bootstrap. The other way is the Bootstrap classes, such as `d-none d-lg-block` which would make something `display: none` by default and `display: block` on large or greater screen sizes https://getbootstrap.com/docs/4.0/utilities/display/

client/src/components/Header/Header.scss
```diff
@@ -1,37 +1,86 @@
...
+        padding: 2rem;
+    }
 
-    .left {
-        flex: 20%;
+    .header-icon {
+        object-fit: cover;
```
benvogler: Moved from a style attribute to the scss

client/src/components/Header/Header.scss
```diff
@@ -1,37 +1,86 @@
...
     }
 
-    .center {
-        flex: 60%;
+    .desktop-links {
+        display: none;
+        @include media-breakpoint-up(lg) {
+            flex: 20%;
+            display: inline-block;
+            margin-left: 2rem;
```
benvogler: There were a lot of places where percents were being used for things like padding and margin, please use rem for that instead https://webdesign.tutsplus.com/tutorials/why-you-should-be-using-rem-based-layouts--cms-27828

client/src/components/Header/Header.scss
```diff
@@ -1,37 +1,86 @@
...
+    }
+
+    .cheeseburger-menu {
+        ul {
+            padding-inline-start: 2rem;
+            padding-right: 2rem;
+            margin: 2rem 0 0;
+            li {
+                list-style-type: none;
+            }
```
benvogler: I fixed up the spacing of the hamburger menu and hid the `li` dots

client/src/components/Header/SearchBar/LocationSearch/LocationSearch.js
```diff
@@ -91,7 +90,7 @@ export default class LocationSearch extends React.Component {
         };
 
         return (
-            <div className="form-control-sm p-0 position-relative" style={{width: width, marginRight: margin, marginLeft: margin, marginBottom: barBottomSpacing}}>
```
benvogler: This stuff is now all directly applied in `Header.scss` so we don't need to pollute the javascript with style information

client/src/components/Header/SearchBar/SearchBar.js
```diff
@@ -55,31 +51,24 @@ class SearchBar extends React.Component {
             d: distanceFilter,
             address: locationAddress
         }, this.props.history);
-        if(typeof this.props.closeModal == 'function') {
-            this.props.closeModal();
+        if(typeof this.props.onCloseModal == 'function') {
```
benvogler: This follows the React naming convention for event handlers

client/src/components/Header/SearchBar/TextSearch/TextSearch.js
```diff
@@ -31,16 +31,15 @@ export default class TextSearch extends React.Component {
 
     render() {
         const { searchPlaceholder, searchQuery } = this.state;
-        const { className, onKeyDown, width, barBottomSpacing } = this.props;
+        const { className, onKeyDown } = this.props;
         return (
-            <div className="position-relative" style={{width: width, marginBottom: barBottomSpacing}}>
+            <div className="text-search position-relative">
```
benvogler: Since I needed to target this element from `Header.scss` I just added a class name to this and did the same for other cases such as `location-search` and `search-button`

client/src/components/Header/SearchBar/TextSearch/TextSearch.js
```diff
@@ -31,16 +31,15 @@ export default class TextSearch extends React.Component {
...
                 <input type="text"
                        value={searchQuery}
                        onChange={(event) => this.handleChange(event)}
                        placeholder={searchPlaceholder}
                        className={className}
-                       onKeyDown={onKeyDown}
-                       style={{width: '100%'}}/>
```
benvogler: I got rid of all these width: 100%'s because they're usually overkill and can be handled better by making the parent `display: flex`

client/src/components/Monument/Details/Address/Address.scss
```diff
@@ -1,6 +1,3 @@
-@import "../../../../../node_modules/bootstrap/scss/functions";
```
benvogler: This is unrelated to your PR, I just noticed I was importing these in some places where they don't need to be. I also have realized that these imports aren't getting deduplicated causing lots of duplicate bootstrap css... hopefully that's just in the dev version but if not we will need to figure out how to avoid this or else our compiled css size will blow up

benvogler:
The failing travis build is unrelated to my changes, not sure how this happened but there's lots of failing tests...

njd9419:
Seems the CI build failed?

client/src/components/Header/Header.js
njd9419: What does `&& <>` do?

njd9419:
Looks good, I just had one small question.

Also, the build is failing because there were unit tests that depended on the year being 2019. That was fixed when #71 was merged to master.

client/src/components/Header/Header.js
benvogler: This is borrowed from @bxsmith007, it makes an empty tag I guess. Nifty way of doing conditional rendering cleanly

njd9419:
Conflicts.

benvogler:
Branch is now pointed at master. Refer to c055fe603af56f576286f32bbb4f74b020e2b182 if you want to see my changes only. I'll fix the conflict and merge tomorrow

PR #81, opened 2020-01-19
New MonumentTag Entity Implementation

njd9419:
This PR refactors how the Monument-Tag Many-to-Many relationship works. Instead of the `MonumentTag` entity being implicit via a `@ManyToMany` annotation, it is now an explicit relationship using 2 `@OneToMany`s on the `Monument` and `Tag` and 2 corresponding `@ManyToOne`s on the `MonumentTag` entity.

There were many things that had to change because of this, such as service methods, queries, fields on Model classes, and unit tests. I **_believe_** I got every area impacted. If you know something I missed, please let me know.

Of course, this was a rather large change and could have pretty negative effects if done incorrectly. I was pretty fearful of that and did regression on the entire application to try and catch any issues. Feel free to pull down this branch and do your own testing if you wish.

These are the areas that I did regression testing on:

- **Initial Data Load:**
-- Both branches loaded 659 Monuments and 571 Tags into the database with 0 invalid rows in the CSV.
- **Searching M+Ms:**
-- I searched for the text "monument", which had both branches return 226 results. I checked the order of the first 5 results. Then, I sorted by newest first and oldest first, confirming the ordering there as well. Then, I added a Tag filter of "Men", which both branches returned 86 results. Then, I added a Material filter of "Stone", where both branches returned 24 results. I then cleared those filtered and spot checked 2 other Tags, both returning the expected number of results.
- **View M+Ms:**
-- I opened the same Monument record on both branches, confirmed that it had the same Tags, same Nearby Monuments in the same order and the same Related Monuments in the same order. I also double-checked that clicking a Tag from this page correctly searched the Tag.
- **Create M+M:**
-- I checked that creating a single M+M using the UI created the result appropriately.
- **Bulk-Create M+Ms:**
-- I confirmed that using the same CSV file on both branches created the same number of Monuments.
- **Bulk-Create M+Ms with Images:**
-- I confirmed that using the same .zip file on both branches created the same number of Monuments.

Once again, sorry this is such a large PR.


src/main/java/com/monumental/controllers/TagController.java
```diff
@@ -7,7 +7,6 @@
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
-import java.security.InvalidParameterException;
```
njd9419: This was just me cleaning up unused imports.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -320,24 +333,49 @@ private void getRelatedRecords(List<Monument> monuments, String fieldName) {
...
+            }
+
+            for (Monument monument : monuments) {
+                try {
+                    Monument.class.getDeclaredMethod("set" + capitalizedFieldName, Set.class).invoke(monument, map.get(monument.getId()));
+                } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
+                    System.err.println("Invalid field name: " + fieldName);
+                    System.err.println("Occurred while trying to use setter: set" + capitalizedFieldName);
+                    e.printStackTrace();
+                }
```
njd9419: We could remove this special case since it's never used, but we'd need to re-add it in the future if we need it.

src/test/java/com/monumental/services/unittest/MonumentServiceUnitTests.java
```diff
@@ -1,7 +1,6 @@
 package com.monumental.services.unittest;
 
 import com.monumental.services.MonumentService;
-import com.monumental.util.csvparsing.BulkCreateResult;
```
njd9419: Cleaning up unused imports.

src/main/java/com/monumental/services/TagService.java
```diff
@@ -49,6 +51,7 @@
 
     /**
      * Safely create a new tag without duplication
+     * This method should be used ANYTIME you wish to associated a Tag with a Monument
```
benvogler:
```suggestion
     * This method should be used ANYTIME you wish to associate a Tag with a Monument
```

njd9419:
@benvogler Like I said, I went through regression as best I could. And yeah, we'll definitely notice if something is off pretty quickly.

PR #83, opened 2020-01-24
Refactor usages of forEach and single quotes in JSX

benvogler:
This PR refactors all but one of our usages of `forEach` to either use `for (let member of array) {` or [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)  if applicable.

Main reason for this was to fix the error on monument pages when they have no tags, but I also replaced other usages of it that worked because they either could be done more easily with `map` or because I prefer `for of` syntax

I also replaced usages of single quotes in JSX with double quotes because it fits the HTML style better and I want us to be consistent with our usage

### DIFF
--- a/client/src/components/CreateForm/CreateForm.js
+++ b/client/src/components/CreateForm/CreateForm.js
@@ -219,14 +219,14 @@ export default class CreateForm extends React.Component {
         inscription.isValid = true;
         inscription.message = '';
 
-        references.forEach(reference => {
+        for (let reference of references) {
             reference.isValid = true;
             reference.message = '';
 
             if (resetValue) {
                 reference.value = '';
             }
-        });
+        }
 
         materials.isValid = true;
         materials.message = '';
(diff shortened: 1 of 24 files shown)


client/src/components/CreateForm/CreateForm.js
```diff
@@ -519,57 +519,53 @@ export default class CreateForm extends React.Component {
...
-                    <Form.Control.Feedback type='invalid'>{reference.message}</Form.Control.Feedback>
-                </div>
-            );
-        });
+        const referenceInputs = references.map((reference, index) => (
+            <div className="reference-container" key={index}>
+                <Form.Label>Reference:</Form.Label>
+                <Form.Control
+                    type="text"
+                    name={'reference-' + index}
```
njd9419: Isn't this considered JSX?

benvogler: Yes but it's within curly brackets so it's JavaScript. I only replaced ones that are just strings since that's more like regular html

PR #84, opened 2020-02-05
Field mapping V1

benvogler:
This PR includes a *mostly* functional PR

The not functional part is that during record creation the request will timeout if it runs over 2 minutes. I'm working on a V2 that will fix that, but since the records are actually created even if the request times out, I want to get this on the VM for tomorrow


client/src/actions/bulk.js
```diff
@@ -0,0 +1,69 @@
...
+        payload: result
+    };
+}
+
+function error(action, error) {
+    return {
+        type: action === actions.validate ? BULK_VALIDATE_MONUMENTS_ERROR : BULK_CREATE_MONUMENTS_ERROR,
+        error: error
+    };
+}
```
njd9419: I like this format when we have multiple action types but with the same basic functionality.

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -16,58 +16,154 @@ export default class BulkCreateForm extends React.Component {
...
+                        fileUpload.errorMessage = 'Your .zip file contained multiple .csv files. Please only upload one .csv file at a time.';
+                        fileUpload.csv = null;
+                        fileUpload.zip = null;
+                        fileUpload.images = [];
+                        fileUpload.isValid = false;
+                        break;
+                    }
+                    fileUpload.csv = content.files[fileName];
+                } else {
+                    fileUpload.images.push(content.files[fileName]);
```
njd9419: I haven't finished reading the PR, but is assuming that all other files are valid image files going to be an issue somewhere? (Maybe I'll answer my own question as I keep reading).

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -16,58 +16,154 @@ export default class BulkCreateForm extends React.Component {
...
         }
         else {
             event.target.value = '';
             alert('Invalid file type submitted');
         }
     }
 
-    handleMoreInformationClick() {
-        const { showingMoreInformation } = this.state;
+    async readCSVHeaders() {
```
njd9419: Do you think we'll ever need to do any other CSV reading on the front-end and should extract some of this to a `util` function?

client/src/components/BulkCreateForm/BulkCreateForm.js
benvogler: It only really matters if a row in the csv lists that filename as an image. Then the backend will check if it's the right file type.

client/src/components/BulkCreateForm/BulkCreateForm.js
benvogler: I don't think we have any other use cases for it at the moment.

client/src/actions/bulk.js
benvogler: I just thought of an even better way of doing this:
```
const actions = {
    validate: {
        pending: BULK_VALIDATE_MONUMENTS_PENDING,
        success: BULK_VALIDATE_MONUMENTS_SUCCESS,
        error: BULK_VALIDATE_MONUMENTS_ERROR
    },
    create: // etc.
};

...

function error(action, error) {
    return {
        type: action.error,
        error: error
    }
}
...
error(actions.validate, error);
```

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -76,164 +172,298 @@ export default class BulkCreateForm extends React.Component {
...
+        fields.forEach(field => field.selected = false);
+
+        onResetForm();
+        this.setState({
+            fileUpload,
+            fileUploadInputKey,
+            fields,
+            showFieldMapping: false,
+            mapping: []
+        });
```
njd9419: Any reason for calling `OnResetForm()` before `this.setState()`?

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -76,164 +172,298 @@ export default class BulkCreateForm extends React.Component {
...
-        let formIsValid = true;
-
-        /* File Upload Validation */
-        /* A File upload is required */
-        if (fileUpload.file.size === undefined) {
-            fileUpload.isValid = false;
-            fileUpload.errorMessage = 'A file is required';
-            formIsValid = false;
-        }
+        return (<>
```
njd9419: What does returning the `<>` empty tag at the beginning do?

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -76,164 +172,298 @@ export default class BulkCreateForm extends React.Component {
...
-
-        let form = {
-            file: fileUpload.file
-        };
-
-        if (fileUpload.fileType === '.zip') {
-            onZipSubmit(form);
+        const { validationResult } = this.props;
+
+        // TODO: Handle general errors
```
njd9419: 👀

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -76,164 +172,298 @@ export default class BulkCreateForm extends React.Component {
...
+                    </Button>
+                </Card.Footer>
+            </>);
         }
-    }
 
-    handleFeedbackModalClose() {
-        this.resetForm(true);
-    }
+        console.log({errorCount, warningCount});
```
njd9419: Minor: Remove debug print.

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -76,164 +172,298 @@ export default class BulkCreateForm extends React.Component {
...
-        const moreInformationLink = (
-            <div className="more-information-link"
-                 onClick={() => this.handleMoreInformationClick()}>
-                Show More Information
-            </div>
-        );
+        let errorString = [];
+        if (errorCount > 0) errorString.push(errorCount + ' error' + (errorCount > 1 ? 's' : ''));
+        if (warningCount > 0) errorString.push(warningCount + ' warning' + (warningCount > 1 ? 's' : ''));
+        console.log(errorString);
```
njd9419: Minor: Remove debug print.

client/src/components/BulkCreateForm/BulkCreateForm.js
benvogler: No, don't think it really matters. `onResetForm` just tells the parent to update the props it passes down into this component

client/src/components/BulkCreateForm/BulkCreateForm.js
benvogler: Yea basically I added an option to the csv validation where it can pass in general errors instead of line-based errors, but it's never used right now. I was expecting some use cases to come up where it might but they really didn't. The only thing I can think of right now is IOExceptions and maybe if it can't connect to S3 or something

client/src/components/BulkCreateForm/BulkCreateForm.js
benvogler: When you make a JSX value it must be only one root element. You can't have:
```
return (
    <div/>
    <div/>
);
```

Originally how I would get around this was
```
return (
    <div>
        <div/>
        <div/>
    </div>
)
```

But in this use case I need `Card.Body` and `Card.Footer` without being wrapped in another element. As I learned from Ben's PR, you can actually make empty elements in react. So, by wrapping these two in an empty element I can return both of them as root level elements, basically, and avoid messing up Bootstrap

njd9419: That's nice.

client/src/components/BulkCreateForm/MoreInformation/MoreInformation.js
```diff
@@ -1,258 +0,0 @@
-import React from 'react';
```
njd9419: Maybe it'll be useful (I guess we can ask Dr. Decker about this) to have some kind of explanation still about Bulk Upload is and how to use it?

client/src/components/Search/Filters/Filters.js
```diff
@@ -57,6 +57,7 @@ class Filters extends React.Component {
                 this.handleFilterChange('decade', value);
                 break;
             case 'range':
+            default:
```
njd9419: Thank god, React was complaining about this forever.

client/src/utils/api-util.js
```diff
@@ -34,14 +34,15 @@ export async function get(url) {
  * Send a POST request to the specified url with the specified data
  * @param url - URL to send the POST to
  * @param data - JSON data to send to the specified URL
+ * @param contentType
```
njd9419: Minor: Missing `@param` description.

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -192,25 +194,38 @@ public Monument updateMonument(@PathVariable("id") Integer id, @RequestBody Monu
     }
 
     /**
-     * Create many Monuments based on the specified CSV file contents
-     * @param csvContents - List of Strings, where each String is a CSV row
-     * @return BulkCreateResult - Object containing information about the Bulk Monument Create operation
+     * Validate which rows in the specified .csv (or .csv within .zip) file are valid
+     * @param request - Contains the field mapping and the file to process
+     * @return BulkCreateResult - Object representing the results of the Bulk Monument Create operation
```
njd9419: Minor: Should be changed to something like `Object representing the results of the Bulk Monument Validate operation`

client/src/components/BulkCreateForm/MoreInformation/MoreInformation.js
benvogler: Yea I'm sure she'll want something. It was easier at the time to just wipe it away while I was working. We can come back to this PR and find the old info and work with it

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -629,76 +548,97 @@ public BulkCreateResult bulkCreateMonumentsFromCsv(List<String> csvList, boolean
...
-     */
-    private String preProcessImageForCsvRow(String csvRow, List<String> imageFileNames, ZipFile zipFile) {
-        String imageFileName = CsvMonumentConverter.getImageFileNameFromCsvRow(csvRow);
-
-        // If the uploaded .zip file contains the CSV row's image filename, upload the image to S3 and
-        // set the CSV row's image filename column to the S3 Object URL
-        if (imageFileNames.contains(imageFileName)) {
-            // Get the ZipEntry for the Image
-            ZipEntry imageZipEntry = zipFile.getEntry(imageFileName);
+    public List<Monument> bulkCreateMonuments(List<CsvMonumentConverterResult> csvResults) {
```
njd9419: Minor: Can we get a docstring for this method?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -629,76 +548,97 @@ public BulkCreateResult bulkCreateMonumentsFromCsv(List<String> csvList, boolean
...
+
+                // Insert all of the Materials associated with the Monument
+                List<String> materialNames = result.getMaterialNames();
+                if (materialNames != null && materialNames.size() > 0) {
+                    for (String materialName : materialNames) {
+                        this.tagService.createTag(materialName, Collections.singletonList(insertedMonument), true);
+                    }
+                }
+
+                // TODO: Upload images to S3
```
njd9419: Do we actually still need to do this?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -629,76 +548,97 @@ public BulkCreateResult bulkCreateMonumentsFromCsv(List<String> csvList, boolean
...
+            } catch (DataIntegrityViolationException e) {
+                // TODO: Determine how duplicate "monument_tag" (join table) records are being inserted
+                // These are disregarded for now - the correct tags are still being created
+                System.out.println("Duplicate monument_tag was not created");
```
njd9419: Can we remove this catch since we found the cause?

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -629,76 +548,97 @@ public BulkCreateResult bulkCreateMonumentsFromCsv(List<String> csvList, boolean
...
+            } catch (Exception e) {
+                if (ExceptionUtils.getRootCause(e).getMessage().contains("duplicate key")) {
+                    // TODO: Determine how duplicate "monument_tag" (join table) records are being inserted
+                    // These are disregarded for now - the correct tags are still being created
+                    System.out.println("Duplicate monument_tag was not created");
```
njd9419: Same here.

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -18,159 +19,174 @@
...
-     * Static method for converting a CSV row representing a Monument into a CsvMonumentConverterResult object
-     * @param csvRow - CSV row representing a Monument, as a String
-     * @param fromZipFile - True if the csvRow originated from a .zip file, False otherwise
-     * @return CsvMonumentConverterResult - the CsvMonumentConverterResult that is represented by the CSV row
-     */
-    public static CsvMonumentConverterResult convertCsvRow(String csvRow, boolean fromZipFile) {
-        if (csvRow == null) {
-            return null;
-        }
+    public static List<CsvMonumentConverterResult> convertCsvRows(List<String[]> csvRows, Map<String, String> mapping,
```
njd9419: Minor: Can we get a docstring for this method?

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -18,159 +19,174 @@
...
+                        result.getMaterialNames().addAll(parseTags(value));
+                        break;
+                    case "inscription":
+                        monument.setInscription(value);
+                        break;
+                    case "latitude":
+                        try {
+                            latitude = Double.parseDouble(value);
+                        } catch (NumberFormatException e) {
+                            result.getWarnings().add("Latitude should be a valid number.");
```
njd9419: Reminder of our conversation about making this a warning and an error.

njd9419:
Failing build.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -603,38 +600,49 @@ public MonumentBulkValidationResult validateMonumentCSV(List<String[]> csvList,
...
+                        String objectUrl = this.s3Service.storeObject(
+                                AwsS3Service.imageFolderName + imageFile.getName(),
+                                imageFile
+                        );
+                        Image image = new Image();
+                        image.setUrl(objectUrl);
+                        image.setMonument(insertedMonument);
+                        image.setIsPrimary(true);
+                        insertedMonument.getImages().add(image);
+                        this.monumentRepository.save(insertedMonument);
```
njd9419: Why save the Monument everytime?

src/main/java/com/monumental/services/AwsS3Service.java
```diff
@@ -24,50 +29,77 @@
...
+        } catch (AmazonServiceException e) {
+            System.out.println("Error attempting to access S3 Bucket: " + bucketName + " and Object: " + objectKey);
+            System.out.println(e.getErrorMessage());
+            return true;
+        }
+    }
+
+    public static String generateUniqueKey(String objectKey) {
+        if (isUniqueKey(objectKey)) return objectKey;
+        else {
```
njd9419: No need for `else` here.

src/main/java/com/monumental/services/AwsS3Service.java
```diff
@@ -24,50 +29,77 @@
...
+            String checkRegex = "/\\([0-9]*\\)/$";
+            String captureRegex = "/\\(([0-9]*)\\)/$";
+            // If the key already ends with "(1)", for example
+            if (objectKey.contains(checkRegex)) {
+                Integer number = Integer.parseInt(Pattern.compile(captureRegex).matcher(objectKey).group(0)) + 1;
+                objectKey = objectKey.replaceAll(checkRegex, "(" + number + ")");
```
njd9419: What if someone uploads an image with the name `image (1).png` but we already have `image (2).png` in S3?

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -18,159 +20,157 @@
...
+                                result.getWarnings().add("Could not find image in .zip file. File may be missing or named incorrectly.");
+                            } else {
+                                try {
+                                    result.getImageFiles().add(ZipFileHelper.convertZipEntryToFile(zipFile, imageZipEntry));
+                                } catch (IOException e) {
+                                    result.getErrors().add("Failed to read image file from .zip");
+                                }
+                            }
+                        } else {
+                            result.getWarnings().add("Cannot upload images with a .csv file. You must package your .csv and your images into a .zip file and upload it.");
```
njd9419: Maybe we should just add this at the top of the page instead of displaying it as a warning for each row.

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -18,159 +20,157 @@
...
-                    if (fromZipFile) {
-                        newImage.setUrl(value);
-                    }
-                    // Otherwise the csvRow came from a CSV file, meaning we have to transform the filename into
-                    // S3 Object URL
-                    else {
-                        newImage.setUrl(AwsS3Service.getObjectUrl(AwsS3Service.imageBucketName, AwsS3Service.imageFolderName + value));
-                    }
+    private static List<String> parseTags(String value) {
+        // Split on commas in-case there are more than one Material in the column
```
njd9419: Minor: Should be `...more than one Tag in the column`

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -443,59 +334,8 @@ public static String cleanTagName(String tagName) {
...
-
-            if (i != (csvRowArray.length - 1)) {
-                newCsvRowBuilder.append(",");
-            }
-        }
-
-        return newCsvRowBuilder.toString();
+        if (tagName.length() > 0) {
+            return tagName.substring(0, 1).toUpperCase() + tagName.substring(1);
+        } else return null;
```
njd9419: Could remove the `else` here.

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverterResult.java
```diff
@@ -21,14 +18,15 @@
 
     private Monument monument;
 
-    private List<String> tagNames;
+    private Set<String> tagNames = new HashSet<>();
 
-    private List<String> materialNames;
+    private List<String> materialNames = new ArrayList<>();
```
njd9419: Should we make this a `Set` too just in-case?

src/main/java/com/monumental/util/csvparsing/MonumentBulkValidationResult.java
```diff
@@ -0,0 +1,49 @@
...
+        return this.results;
+    }
+
+    public void setResults(Map<Integer, CsvMonumentConverterResult> results) {
+        this.results = results;
+    }
+
+    public String getError() {
+        return this.error;
+    }
```
njd9419: This getter is not used. I'm guessing this is for the "general" errors you talked about? Do we still want this?

src/main/java/com/monumental/util/csvparsing/ZipFileHelper.java
```diff
@@ -75,17 +64,23 @@ public static ZipFile convertMultipartFileToZipFile(MultipartFile multipartFile)
      * @throws IOException - If any I/O errors occur when reading from the ZipFile or ZipEntry
      */
     public static File convertZipEntryToFile(ZipFile zipFile, ZipEntry zipEntry) throws IOException {
+
```
njd9419: Minor: No reason for newline here.

src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
```diff
@@ -45,61 +50,91 @@
...
-        assertNull(this.monumentService.bulkCreateMonumentsFromCsv(null, false, null, null));
+    private MonumentBulkValidationResult validateCSV(String csvRows) {
+        try {
+            List<String[]> csvList = parseCSVString(csvRows);
+            return this.monumentService.validateMonumentCSV(csvList, mapping, null);
+        } catch (IOException e) {
+            e.printStackTrace();
+            fail("An IOException occurred");
+            return new MonumentBulkValidationResult();
+        }
```
njd9419: What's the reason for these two `static` methods in here?

njd9419:
Done with my first round of review.

I have some requests for comments as well as some questions about some things.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -629,76 +474,111 @@ public BulkCreateResult bulkCreateMonumentsFromCsv(List<String> csvList, boolean
...
-                fileToUpload.delete();
-                // Set the CSV Row's image filename column to the Object URL
-                return CsvMonumentConverter.setImageFileNameOnCsvRow(csvRow, objectUrl);
-            } catch (IOException e) {
-                e.printStackTrace();
-                return CsvMonumentConverter.setImageFileNameOnCsvRow(csvRow, "");
+    public List<Monument> bulkCreateMonuments(List<CsvMonumentConverterResult> csvResults) {
+        List<Monument> monuments = new ArrayList<>();
+        for (CsvMonumentConverterResult result : csvResults) {
+            int index = csvResults.indexOf(result);
```
njd9419: This is unused.

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
benvogler: I think we decided to leave this, right? Because as a warning, it allows you to still insert if you have a valid address whereas if I make it an error then you would not be able to. It becomes a warning if you have no address since then you have no coordinates or address

src/main/java/com/monumental/services/AwsS3Service.java
benvogler: This now is in a loop until it finds a unique key

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
benvogler: V2

src/main/java/com/monumental/util/csvparsing/MonumentBulkValidationResult.java
benvogler: It'll be used in v2

src/test/java/com/monumental/services/integrationtest/MonumentServiceIntegrationTests.java
benvogler: Basically I had a bunch of code copy and pasted in the tests so I broke them out into these functions

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
njd9419: Yeah, you're right.

PR #85, opened 2020-02-05
Update Single M+M

njd9419:
This PR adds:

1. The ability to update an existing Monument through the UI.
2. Fixes the `CreatedDate` and `LastModified` fields on `Model`.

**Updating an existing Monument through the UI:**
The experience for this is that the user goes to the Update Page for a particular Monument. The Page has the familiar Create Form, but pre-filled with the current values of the Monument. Here, the user is able to modify the fields as well as delete/add new things where appropriate. Once completed, the user is presented with a modal where they can review their changes before confirming the update.

Screenshots:

The CreateForm now pre-filled with the Monument's attributes:
image

image

Review modal after changes are made:
image

Expanded review modal to show unchanged attributes:
image

**Fixing CreatedDate and LastModifiedDate:**
Ever since we switched to using `AbstractAuditable`, these two attributes where being inserted as `NULL` for new records.

I re-added the annotations we had in `Model` before the switch to `AbstractAuditable` and that fixed the issue.


client/src/utils/api-util.js
```diff
@@ -95,6 +95,38 @@ export async function postFile(url, file) {
     else return res;
 }
 
+/**
+ * Send a PUT request to the specified url with the specified data
+ * @param url - URL to send the PUT to
+ * @param data - JSON data to send to the specified URL
+ */
+export async function put(url, data) {
```
benvogler: You should abstract the code from `POST` into a generic function that accepts the method type, no need to copy and paste. Also there will be conflicts with my PR because I made content type configurable

client/src/utils/api-util.js
```diff
@@ -132,4 +164,51 @@ export default async function uploadImagesToS3(images) {
...
+}
+
+/**
+ * Helper function to get the S3 Object Key for an Image using the specified Object URL
+ * @param objectUrl - The Object URL to use to parse the Image Object Key
+ */
+function getS3ImageObjectKeyFromObjectUrl(objectUrl) {
+    const objectUrlArray = objectUrl.split('/');
+    return s3ImageBucketFolderName + objectUrlArray[objectUrlArray.length - 1];
+}
```
benvogler: One of the bugs I fixed was that we urlencode the key before storing it in the image table. You should make sure this still works after that's merged

client/src/utils/api-util.js
```diff
@@ -132,4 +164,51 @@ export default async function uploadImagesToS3(images) {
...
+
+/**
+ * Helper function to get the Image name using the specified S3 Object URL
+ * @param objectUrl - The Object URL to use to parse the Image name
+ */
+export function getS3ImageNameFromObjectUrl(objectUrl) {
+    const objectUrlArray = objectUrl.split('/');
+    return objectUrlArray[objectUrlArray.length - 1];
 }
```
benvogler: Same here

client/src/actions/update-monument.js
```diff
@@ -0,0 +1,85 @@
...
+        payload: updatedMonument
+    };
+}
+
+function updateMonumentError(error) {
+    return {
+        type: UPDATE_MONUMENT_ERROR,
+        error: error
+    };
+}
```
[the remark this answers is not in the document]
njd9419: I'll do that, but there's not a nice way to abstract out the functions into `doAction` like you did unfortunately since the actions use different HTTP methods. It'd end up being more complex than good.

client/src/utils/api-util.js
njd9419: Going to double check that this and the other function still work now.

njd9419: Alright, made a fix to this function. Had to decode the URL before splitting it into an array. Going to check the other one now.

client/src/utils/api-util.js
njd9419: Same here, the fix was to decode the URL before splitting it.

client/src/utils/api-util.js
```diff
@@ -1,107 +1,105 @@
...
+    }
 }
 
 /**
  * Uploads the specified images to the monument-images S3 bucket, inside the images/ folder
  * @param images - List of images to upload to S3
  * @returns {Promise<[]>} - Promise that when awaited, returns a List of S3 Object keys for the uploaded images
  */
-export default async function uploadImagesToS3(images) {
+export async function uploadImagesToS3(images) {
```
benvogler: I missed this when it was originally written, but we can't use secret keys on the frontend like this as you could theoretically search the minified production code and find it. We will have to use STS to send temporary credentials to the client from the server: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/prog-services-sts.html

We can do this in a separate PR. We need to remember to rotate our security keys once it's done as these ones are potentially exposed

client/src/utils/api-util.js
```diff
@@ -133,4 +131,53 @@ export default async function uploadImagesToS3(images) {
     }
 
     return imageUrls;
+}
+
+export async function deleteImagesFromS3(imageUrls) {
```
benvogler: Same here about secret key

client/src/utils/api-util.js
benvogler: It might even be better if we just upload the file to our server and let it handle all image uploads/deletes

client/src/utils/api-util.js
```diff
@@ -1,107 +1,105 @@
...
-
-                throw Error(errorMessage.message);
-            }
-            return res;
-        })
-        .then(res => res.json())
-        .catch(err => error = err);
-    if (error || res.error) throw(error || res.error);
-    else return res;
+export async function post(url, data) {
```
benvogler: I don't know why I didn't recommend this earlier, but we've basically just recreated [https://www.npmjs.com/package/request-promise-native](request-promise-native) so I'm going to refactor everything to use that

benvogler: never mind I forgot how infuriating it is that this package can't use relative paths, so we'd have to store an environment variable with the server's domain name and I ain't about that

benvogler: We still need to accept headers on these since we don't use JSON in all our requests

client/src/utils/api-util.js
njd9419: I'm all for us removing all S3 Image uploading from the Front-End. That way it all happens in one place again. We'll have to remember to address that in a PR soon.

client/src/utils/api-util.js
njd9419: I added a `contentType` param to `post()`. There was already one on `put()`. I think we're good now, right?

PR #87, opened 2020-02-08
Async Bulk Monument Creation

benvogler:
This is the V2 bulk monument creation which will primarily add a progress bar during creation. It'll also include a couple small tweaks I didn't get into the last version, such as limiting the height of the card when we show the image files found in the .zip


src/main/java/com/monumental/config/AppConfig.java
```diff
@@ -6,11 +6,13 @@
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.support.ResourceBundleMessageSource;
+import org.springframework.scheduling.annotation.EnableAsync;
 
 /**
  * Class used to configure Beans for the application
  */
 @Configuration
+@EnableAsync
```
benvogler: @njd9419 This allows us to use `@Async` on service methods

src/main/java/com/monumental/services/AsyncMonumentService.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.File;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * This class is the async counterpart to the MonumentService, where any asynchronous tasks
+ * should be kept.
+ */
+@Service
+public class AsyncMonumentService extends AsyncService {
```
benvogler: @njd9419 I split this into a separate service for clarity, but it could just as easily work within the `MonumentService` if it extended `AsyncService`

src/main/java/com/monumental/util/async/AsyncJob.java
```diff
@@ -0,0 +1,45 @@
+package com.monumental.util.async;
+
+import java.util.concurrent.CompletableFuture;
+
+public class AsyncJob {
```
benvogler: The `AsyncJob` class wraps the `CompletableFuture` class from the Java Concurrency API. `AsyncJob` gives us the ability to give jobs an Id, store a `CompletableFuture`, and have a progress value (double) associated with the `CompletableFuture

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -211,21 +217,53 @@ public MonumentBulkValidationResult validateMonumentCSV(@ModelAttribute BulkCrea
...
+         * it away currently because the AsyncJob must be passed to the CompletableFuture method, and the CompletableFuture
+         * must be passed to the AsyncJob, making it difficult to do so dynamically
+         */
+        AsyncJob job = this.asyncMonumentService.createJob();
+        job.setFuture(this.asyncMonumentService.bulkCreateMonuments(
+                job,
                 new ArrayList<CsvMonumentConverterResult>(validationResult.getValidResults().values())
-        );
+        ));
+        return job;
```
benvogler: Here we:

* Create an empty `AsyncJob`. At this point all it has is a unique Id, and is stored within our `AsyncMonumentService` so that we can retrieve it by Id in later requests.
* Run the `bulkCreateMonuments` method, which returns an **incomplete** `CompletableFuture`. Think of this as an unresolved promise in Javascript.
* Pass the `AsyncJob` into the `bulkCreateMonuments` method so that it can update its progress as it runs

So, what gets returned here looks like this: 
```
{
    "id": 1,
    "future": {
        "done": false,
        "cancelled": false,
        "completedExceptionally": false,
        "numberOfDependents": 0
    },
    "progress": 0.0
}
```

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -211,21 +217,53 @@ public MonumentBulkValidationResult validateMonumentCSV(@ModelAttribute BulkCrea
...
+    }
+
+    /**
+     * Check the progress of a create bulk monuments job
+     * @param id - Id of the job to check
+     * @return AsyncJob - Object containing the Id of the job and the current value of the Future object
+     */
+    @GetMapping("/api/monument/bulk/create/progress/{id}")
+    public AsyncJob getBulkCreateMonumentJob(@PathVariable Integer id) {
+        return this.asyncMonumentService.getJob(id);
```
benvogler: Here, we retrieve the started job by its Id, so that the front end can check if it's done yet. This is something that will be run on an interval by the UI until it sees that the job has completed.

What gets returned here looks something like:
```
{
    "id": 1,
    "future": {
        "done": false, // Or true if it's done
        "cancelled": false,
        "completedExceptionally": false,
        "numberOfDependents": 0
    },
    "progress": 0.65 // Or some other percent, or 1.0 if it's done (although the "done" field should be used to check this)
}
```

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -211,21 +217,53 @@ public MonumentBulkValidationResult validateMonumentCSV(@ModelAttribute BulkCrea
...
+     * @param id - Id of the job to get the result of
+     * @return List<Monument> - The monuments created
+     * @throws ExecutionException - Can be thrown by Java if the future encountered an exception
+     * @throws InterruptedException - Can be thrown by Java if the future encountered an exception
+     */
+    @GetMapping("/api/monument/bulk/create/result/{id}")
+    @SuppressWarnings("unchecked")
+    public List<Monument> getBulkCreateMonumentJobResult(@PathVariable Integer id)
+            throws ExecutionException, InterruptedException {
+        return (List<Monument>) this.asyncMonumentService.getJob(id).getFuture().get();
```
benvogler: This *should* be called after the frontend *knows* that the job is done. Theoretically, you could call it early, in which case Java will wait for the job to finish. I may add a check to this so that Java aborts if it sees that the job isn't done yet, to prevent the thread waiting for the job to finish.

What's returned here is the final result of the operation, in this case the list of created monuments

src/main/java/com/monumental/services/AsyncMonumentService.java
```diff
@@ -0,0 +1,100 @@
...
+    AwsS3Service s3Service;
+
+    /**
+     * Insert validated monument CSV rows and related objects
+     * @param job - The AsyncJob to report progress to
+     * @param csvResults - The validated CSV rows, converted into monuments
+     * @return List of inserted monuments
+     */
+    @Async
+    public CompletableFuture<List<Monument>> bulkCreateMonuments(AsyncJob job, List<CsvMonumentConverterResult> csvResults) {
```
benvogler: Since this is marked as `@Async`, Spring will do a bunch of proxy magic to make this run in a separate thread. No need for us to mess with that stuff 🙌

src/main/java/com/monumental/services/AsyncMonumentService.java
```diff
@@ -0,0 +1,100 @@
...
+                }
+                if (encounteredS3Exception) {
+                    result.getErrors().add("An error occurred while uploading image(s). Try uploading the images again later.");
+                }
+            }
+
+            // Report progress
+            if (i != csvResults.size()) {
+                job.setProgress((double) (i + 1) / csvResults.size());
+            }
```
benvogler: Since the `AsyncJob` is passed into this method, we can report the progress. In this case, we report the percentage of monuments done being created. I'm ignoring the final monument here so that the progress is not updated to 100% until the final `saveAll` is run below

src/main/java/com/monumental/services/AsyncService.java
```diff
@@ -0,0 +1,28 @@
...
+import com.monumental.util.async.AsyncJob;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This abstract class should be extended by asynchronous services that need to maintain
+ * an internal job queue in order to report progress and results of AsyncJobs as they run
+ */
+public abstract class AsyncService {
```
benvogler: This abstract class should be extended by async services and basically just maintains an internal map of jobs using their id, so that they can be retrieved later to check the progress or get the final result

client/src/actions/bulk.js
```diff
@@ -45,15 +44,33 @@ function buildFormData(form) {
...
+                        result = await (await fetch(`/api/monument/bulk/${action}/progress/${jobId}`)).json();
+                        dispatch(pending(action, result.progress));
+
+                        if (result.future.done) resolve();
+                    }, 200);
+                });
+                window.clearInterval(interval);
+
+                result = await (await fetch(`/api/monument/bulk/${action}/result/${jobId}`)).json();
+            }
```
benvogler: This is where we start the job and get progress of it (every 200ms currently), then once the job is done get the final results and send them to the reducer

client/src/pages/MonumentBulkCreatePage/MonumentBulkCreatePage.js
```diff
@@ -53,23 +54,37 @@ class MonumentBulkCreatePage extends React.Component {
...
+                    <Modal.Header className="pb-0">
+                        <Modal.Title>
+                            Bulk Creating Monuments or Memorials
+                        </Modal.Title>
+                    </Modal.Header>
+                    <Modal.Body>
+                        <div className="mb-2">
+                            Please wait while your monuments or memorials are created...
+                        </div>
+                        <ProgressBar now={createProgress * 100}/>
```
benvogler: Thank god for this component.

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -214,9 +219,25 @@ export default class BulkCreateForm extends React.Component {
...
+            'Errors': result.errors.join('\n')
+        }});
+        const csv = 'data:text/csv;charset=utf-8,' + toCSV(results, {fields});
+        const encodedUri = encodeURI(csv);
+        const link = document.createElement("a");
+        link.setAttribute("href", encodedUri);
+        link.setAttribute("download", `Validation Results ${moment().format('YYYY-MM-DD hh:mm')}.csv`);
+        document.body.appendChild(link);
+        link.click();
+    }
```
njd9419: This is snazzy.

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -461,6 +486,9 @@ export default class BulkCreateForm extends React.Component {
                 }
             </Card.Body>
             <Card.Footer className="d-flex justify-content-end">
+                <Button variant="light" className="mr-2" onClick={() => this.downloadCSV(results)}>
```
njd9419: 👏

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -478,4 +506,16 @@ export default class BulkCreateForm extends React.Component {
...
         </>);
     }
+
+    renderCreateResults() {
+        const { createResult } = this.props;
+        return (
+            <Card.Body>
+                <h5>Success!</h5>
+                <div>
+                    {createResult.length} monuments or memorials have been created. I would list them out here, but pretty soon they're just going to be suggestions anyway!
```
njd9419: Got 'em.

src/main/java/com/monumental/services/AsyncJobService.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This service maintains an internal job queue in order to report progress and results of AsyncJobs as they run
+ */
+@Service
+public class AsyncJobService {
+
+    private Map<Integer, AsyncJob> jobs = new HashMap<>();
```
njd9419: Do jobs ever get removed from this `Map`?

njd9419:
Code review done. Overall, things look great. Nice abstractions so that we can build other async operations in the future if we need. Very understandable too.

Just 2 quick questions:
1. I posted above about the jobs `Map`.
2. What's the UI experience like for a very small CSV? Does the progress bar move so fast that it's annoying?

client/src/components/BulkCreateForm/BulkCreateForm.js
benvogler: Thanks stackoverflow

src/main/java/com/monumental/services/AsyncJobService.java
benvogler: No, only when the server restarts, since this is only stored in memory

benvogler:
@njd9419 the bar starts to go up but the modal closes before the bar's animation finished. I'm not really bothered by it enough to fix it

PR #88, opened 2020-02-11
Mobile search page - additional features

bxsmith007:
I forgot to let the group know that I had one more thing to add before merging, so that is my fault. This should hopefully be a quick approval and addition from the changes made in the first Mobile search page PR.


bxsmith007:
Screen Shot 2020-02-11 at 12 08 13 PM

This is what it looks like on mobile. Do you think anything should change in the smaller section, like where "Show" and "Sort by" are?

njd9419:
@bxsmith007 I think the "Show" and "Sort By" are fine where the are, but there should probably be just a bit of whitespace between them so the boxes aren't touching like they are in that screenshot.

njd9419:
Also, aren't all of these commits the ones you already merged in #76 ?

client/src/components/Search/Search.js
```diff
@@ -12,11 +12,6 @@ import * as moment from 'moment';
  */
 export default class Search extends React.Component {
 
-    handlePageChange(page) {
-        const { onPageChange } = this.props;
-        onPageChange(page);
-    }
-
```
benvogler: Why was this deleted? looks like it's still being called in the `Pagination` below

client/src/components/Search/Search.scss
```diff
@@ -3,68 +3,104 @@
...
+  }
+
+  @include media-breakpoint-down(lg) {
+    width: 100%;
+    flex: 0 0 100%;
+    height: 70%;
+  }
+
+  @include media-breakpoint-up(lg) {
+    width: 40rem;
```
benvogler: Why is the width set to `40rem`? This is very unusual

PR #91, opened 2020-02-12
Add Ability to Geocode Addresses

njd9419:
This PR adds the ability to geocode addresses into coordinates using Google Maps.

Before we were only able to reverse geocode coordinates into an address, but we have a need for both.


njd9419:
This is not done, please do not merge. I still need to finish refactoring the logic that was removed from `MonumentListener`.

njd9419:
This is all set for review.

PR #95, opened 2020-02-21
Sort Tags Alphabetically

njd9419:
This PR adds sorting to the `Tags` component so that it displays them alphabetically by name.

### DIFF
--- a/client/src/components/Tags/Tags.js
+++ b/client/src/components/Tags/Tags.js
@@ -12,7 +12,7 @@ export default class Tags extends React.Component {
         if (!tags) return (<div/>);
         return (
             <div className="tags">
-                {tags.map(tag => {
+                {tags.sort((a, b) => (a.name > b.name) ? 1 : -1).map(tag => {
                     return (
                         <Tag key={tag.id} name={tag.name} isMaterial={tag.isMaterial} selectable={selectable} onSelect={value => onSelect(value, tag)} selectedIcon={selectedIcon} selected={tag.selected}/>
                     );
--- a/src/main/java/com/monumental/controllers/TagController.java
+++ b/src/main/java/com/monumental/controllers/TagController.java
@@ -7,6 +7,8 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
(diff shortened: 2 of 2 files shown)


benvogler:
I think you should do this on the backend as well since there are a lot of places we render tags/materials. I would keep the sort on the frontend since it happens after we combine materials and tags.

src/main/java/com/monumental/controllers/TagController.java
```diff
@@ -30,14 +32,19 @@
...
+            tags = this.tagRepository.getAllByName(name);
         } else if (names != null && names.size() > 0) {
-            return this.tagRepository.getAllByNameIn(names);
+            tags = this.tagRepository.getAllByNameIn(names);
         } else {
-            return this.tagRepository.findAll();
+            tags = this.tagRepository.findAll();
         }
+
+        tags.sort(Comparator.comparing(Tag::getName));
```
njd9419: Java 8 has some neat stuff like this.

njd9419:
I added a sort to the `/api/tags` endpoint. I think that should cover 99% of our bases.

src/main/java/com/monumental/controllers/TagController.java
benvogler: I'm trying to decide if this is better or worse than the Javascript equivalent:
```
tags.sort((a, b) => b.getName() - a.getName());
// or if sorting an array of strings
tags.sort();
```

client/src/components/Tags/Tags.js
```diff
@@ -12,7 +12,7 @@ export default class Tags extends React.Component {
         if (!tags) return (<div/>);
         return (
             <div className="tags">
-                {tags.map(tag => {
+                {tags.sort((a, b) => (a.name > b.name) ? 1 : -1).map(tag => {
```
benvogler: speaking of which, you could use that here
```
tags.sort((a, b) => b.name - a.name)
```

njd9419: I'll do that, it's much cleaner.

njd9419: Actually that doesn't seem to work for me.

benvogler: yea I'm dumb, that just returns NaN

PR #96, opened 2020-02-22
Duplicate Monument Detection

njd9419:
This PR adds duplicate Monument detection to Single and Bulk Create operations.

Currently, a "duplicate" is defined as having a similar name (.9 pg_tgrm score) and within .1 miles of the specified coordinates.

Here are some screenshots of the UI interactions:

**Duplicates detected during Single Create:**
image

From here, the user can click on one of those links and be brought to the View Page for the duplicate or continue with their creation if they feel it's not a duplicate.

**Duplicates detected during Bulk Create:**
image

Duplicates detected during Bulk Create are shown as a warning. That way, the user will be notified but can still continue if they wish.


src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -535,7 +540,16 @@ public MonumentBulkValidationResult validateMonumentCSV(List<String[]> csvList,
...
         List<CsvMonumentConverterResult> results = CsvMonumentConverter.convertCsvRows(csvList, mapping, zipFile);
         for (int i = 0; i < results.size(); i++) {
-            monumentBulkValidationResult.getResults().put(i + 1, results.get(i));
+            CsvMonumentConverterResult result = results.get(i);
+
+            List<Monument> duplicates = this.findDuplicateMonuments(result.getMonument().getTitle(),
+                    result.getMonument().getLat(), result.getMonument().getLon(), result.getMonument().getAddress());
+
+            if (duplicates.size() > 0) {
+                result.getWarnings().add("Potential duplicate records detected for this row.");
```
benvogler: I think you should include links to the duplicates. I think you can use relative urls like `/monuments/1` but you might want to use something similar to `publicUrl` in the `UserService` so that links work while using the react dev server as well

build.gradle
```diff
@@ -59,7 +59,7 @@ dependencies {
     implementation group: 'org.opengeo', name: 'geodb', version: '0.8'
     testImplementation('org.springframework.boot:spring-boot-starter-test')
     testImplementation('junit:junit:4.12')
-    testImplementation('org.mockito:mockito-core:2.1.0')
+    testImplementation('org.mockito:mockito-core:2.13.0')
```
njd9419: I had to upgrade our Mockito version because the old version we were using was missing some implementations we needed.

njd9419:
@benvogler Can you take a look at the last 2 commits I pushed?

PR #103, opened 2020-02-26
Add API endpoint security

benvogler:
This is a quick PR to lock down existing API endpoints. Please take a look through all the controller files (even those unchanged) and see if I missed anything.

### DIFF
--- a/src/main/java/com/monumental/controllers/MonumentController.java
+++ b/src/main/java/com/monumental/controllers/MonumentController.java
@@ -8,6 +8,8 @@ import com.monumental.exceptions.InvalidZipException;
 import com.monumental.exceptions.ResourceNotFoundException;
 import com.monumental.models.Monument;
 import com.monumental.repositories.MonumentRepository;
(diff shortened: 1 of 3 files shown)


benvogler:
Something that just occurred to me is that we don't have a way to require email verification on endpoints currently

src/main/java/com/monumental/controllers/MonumentController.java
```diff
@@ -104,6 +109,7 @@ public Monument updateMonument(@PathVariable("id") Integer id, @RequestBody Upda
      * @return BulkCreateResult - Object representing the results of the Bulk Monument Validate operation
      */
     @PostMapping("/api/monument/bulk/validate")
+    @PreAuthorize(Authorization.isPartnerOrResearcher)
```
njd9419: Does using this also ensure that they are authenticated?

benvogler: Yes

njd9419:
Conflicts.

PR #104, opened 2020-02-27
LastModifiedBy/CreatedBy implementation and tests

benvogler:
This is like part 1/4 for the auditing changes that will come in Milestone 8. This one implements the `lastModifiedBy` and `createdBy` so that they're automatically filled in using the session user, if one exists. It also tweaks some of the session handling so that we can get our `User` object without having to do any JPA queries and adds two new test files:

`UserServiceIntegrationTests`, which basically is meant to test that the way I'm running as a certain user within the tests works

`AuditingIntegrationTests`, which is kind of a weird one that doesn't relate to any specific service/model/repository, instead it tests the `AuditingEntityListener` on the `Model` class and checks that `createdDate`/`createdBy`/`lastModifiedDate`/`lastModifiedBy` are all updated appropriately.

The next step for auditing will probably be to see if there's a standard way to have a table of audit logs so that we can display all changes in the admin panel rather than just most recent.

### DIFF
--- a/src/main/java/com/monumental/security/UserAwareUserDetails.java
+++ b/src/main/java/com/monumental/security/UserAwareUserDetails.java
@@ -1,0 +1,65 @@
+package com.monumental.security;
+
+import com.monumental.models.User;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.userdetails.UserDetails;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * This class wraps the standard UserDetails interface, allowing us to store our own User model within the session object
+ */
+public class UserAwareUserDetails implements UserDetails {
+
+    private final User user;
+    private final Collection<? extends GrantedAuthority> grantedAuthorities;
+
+    public UserAwareUserDetails(User user) {
+        this(user, new ArrayList<GrantedAuthority>());
+    }
+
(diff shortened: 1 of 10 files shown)


src/main/java/com/monumental/security/UserAwareUserDetails.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.userdetails.UserDetails;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * This class wraps the standard UserDetails interface, allowing us to store our own User model within the session object
+ */
+public class UserAwareUserDetails implements UserDetails {
```
njd9419: This class is pretty cool.

src/test/java/com/monumental/services/integrationtest/UserServiceIntegrationTests.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    @WithUserDetails(PARTNER)
+    public void getCurrentUser_partner() {
+        User user = this.userService.getCurrentUser();
+        assertEquals(PARTNER, user.getEmail());
+        assertEquals(Role.PARTNER, user.getRole());
+    }
+
+    @Test
+    @WithUserDetails(COLLABORATOR)
```
njd9419: This annotation is pretty cool

PR #106, opened 2020-02-27
Changes for User Testing

njd9419:
This PR:

1. Adds temporary links in the header to Create and Bulk Create
2. Redirects the Suggest a Change Button on the View M+M Page to Update
3. Allows Collaborators to access Bulk Create


client/src/components/Header/Header.js
```diff
@@ -13,7 +13,9 @@ class Header extends React.Component {
     links = [
         {name: 'Home', route: '/', exact: true},
         {name: 'Map', route: '/map'},
-        {name: 'About', route: '/about'}
+        {name: 'About', route: '/about'},
+        {name: 'Create', route: '/create', exact: true, protected: true},
+        {name: 'Bulk Create', route: '/bulk-create', exact: true, protected: true}
```
benvogler: 1. How does this look on small monitor sizes? I don't think there's actually room for this...
2. For Create we can just update the suggest a change button to link to it instead

client/src/components/Header/Header.js
```diff
@@ -52,9 +54,20 @@ class Header extends React.Component {
...
-        );
+        let headerLinks = this.links.map(link => {
+            let navLink = <NavLink to={link.route} exact={link.exact} className="header-link mr-4" activeClassName="active" key={link.name}>{link.name}</NavLink>;
+
+            if (link.protected) {
+                if (session.user) {
+                    return navLink;
+                }
+                else {
+                    return <div/>;
```
benvogler:
```suggestion
                    return null;
```

client/src/components/Monument/SuggestChanges/SuggestChanges.js
```diff
@@ -17,7 +24,7 @@ export default class SuggestChanges extends React.Component {
                 <Card.Body>
                     <p>We pride ourselves on keeping up to date and accurate information.</p>
                     <p>If you think something on this page is incorrect or outdated, please suggest a change. If your change is approved we'll send you an email to let you know, and you will be listed as a contributor on this page!</p>
-                    <Button variant="primary">SUGGEST A CHANGE</Button>
+                    <Button variant="primary" onClick={() => this.handleButtonClick()}>SUGGEST A CHANGE</Button>
```
benvogler: You can just make this a link instead of passing callbacks all the way down. You will need to pass the `monument` down though
```<Link className="btn btn-primary" href={`/update-monument/${monument.id}`}>SUGGEST A CHANGE</Link>```

client/src/components/Header/Header.js
njd9419: I remove the Create link from the Header and moved it to the Suggest a Change Button on the Homepage.

It doesn't look awful on a smaller monitor, the Log Out and My Account buttons kind of slide off the side though.

client/src/components/Monument/SuggestChanges/SuggestChanges.js
njd9419: I don't think we can do that anymore now that the button links to different pages based on if its on the Home page or the View page.

benvogler: Ok. In the long run I would prefer to have it be a `Link` because it allows you to right click and open in new tab whereas buttons do not. But for the purposes of getting things ready for user testing I don't think it needs to be fixed right away

PR #107, opened 2020-02-28
Add Admin Role

njd9419:
This PR is a quick one to add an Admin Role.

The Admin Role is the only Role that can manage other users' roles and cannot have its Role changed.


njd9419:
@benvogler I think I did this correctly, make sure :)

PR #109, opened 2020-03-04
Fix JPA Auditing causing transaction rollbacks

benvogler:
There was a bug with JPA Auditing `UserService.getCurrentUser()` would throw `UnauthorizedException` if not logged in, which `JpaAuditConfiguration.getCurrentAuditor` would catch and return `null`.

Since `UnauthorizedException` extended `RuntimeException`, Spring would automatically rollback the database transaction. The end result is that any endpoint that did not require authentication was unable to use `@Transactional`

The fix was simply to make `UnauthorizedException` extend `Exception` instead of `RuntimeException`.


njd9419:
Compile errors in `UserServiceIntegrationTests`

PR #113, opened 2020-03-24
Front-End Suggestions

njd9419:
This PR:

1. Fixes the front-end so that it uses the appropriate API endpoints for Suggestions
2. Replaces usages of "Create" and "Update" on the front-end with "Suggest" as appropriate
3. Re-adds the proper authorization/authentication for our routes and endpoints that were removed for user testing

This PR does NOT add the UI for viewing, approving and rejecting Suggestions. That is currently in development and coming in my next PR.


client/src/actions/bulk.js
```diff
@@ -34,16 +34,17 @@ function buildFormData(form) {
...
         try {
-            let result = await (await fetch(action.uri + (isAsyncJob ? '/start' : ''), {
+            let result = await (await fetch(action.uri, {
                 method: 'post',
                 body: buildFormData(form)
             })).json();
 
-            if (isAsyncJob) {
+            // TODO: Use this logic when approving a BulkCreateMonumentSuggestion
+            /*if (isAsyncJob) {
```
benvogler: @njd9419 Is inserting a bunch of suggestions not slow? I feel like we still may need this to use AsyncJobs

njd9419: That's a good point that I hadn't considered. I haven't tested it but we have the capability to do it via `AsyncJob` so we might as well. I'll do that in the Suggestions UI PR since I'm going to be doing it for `BulkCreateMonumentSuggestion` approval then too.

PR #114, opened 2020-04-01
Bug Fixes Round 1

benvogler:
* On the server-side, fix the logic for preventing image name duplicates that was causing `(1) (1)` instead of `(2)` (TODO: Remove client-side image uploading directly to S3, instead route through the server so that this logic is used. Suggestions PR is a pre-req of this)
* Allow choosing between coordinates and street address on create/update monument form to make it abundantly clear that we only want one
* Allow coordinates to be more than 6 decimals long
* Fix a bug with `lastModifiedBy` where a User could become their own `lastModifiedBy` causing JPA to refuse to load the User record in ever, basically locking the user out completely. We now have a special check when setting `lastModifiedBy` in the user class to prevent this (it sets it to null for now) but still allow other users to be set as `lastModifiedBy`
* The big one: reset reducer state after page change where appropriate. This fixes a *lot* of bugs across the site and was easy to implement :tada:


client/src/index.js
```diff
@@ -6,10 +6,24 @@ import { Provider } from 'react-redux';
 import { applyMiddleware, combineReducers, compose, createStore } from 'redux';
 import ReduxThunk from 'redux-thunk';
 import Reducers from './reducers';
+import { createBrowserHistory } from 'history';
+import { routerMiddleware, connectRouter } from 'connected-react-router';
 
-const store = createStore(combineReducers(Reducers), {}, compose(applyMiddleware(ReduxThunk)));
```
njd9419, line 10: Redux is no longer thinking 😢

client/src/index.js
```diff
@@ -6,10 +6,24 @@ import { Provider } from 'react-redux';
...
 
-ReactDOM.render(<Provider store={store}><App /></Provider>, document.getElementById('root'));
+const store = createStore(
+    combineReducers({
+        ...Reducers,
+        router: connectRouter(history)
+    }),
+    {},
+    compose(applyMiddleware(
+        ReduxThunk,
```
njd9419, line 21: I stand corrected!!!!!!!!!!! 🎊

client/src/reducers/bulk.js
```diff
@@ -2,6 +2,7 @@ import {
     BULK_CREATE_MONUMENTS_PENDING, BULK_CREATE_MONUMENTS_SUCCESS, BULK_CREATE_MONUMENTS_ERROR,
     BULK_VALIDATE_MONUMENTS_PENDING, BULK_VALIDATE_MONUMENTS_SUCCESS, BULK_VALIDATE_MONUMENTS_ERROR
 } from '../constants';
+import { LOCATION_CHANGE } from "connected-react-router";
```
njd9419, line 5: Minor: Use `'` instead of `"`

client/src/reducers/create.js
```diff
@@ -1,4 +1,5 @@
 import { CREATE_MONUMENT_PENDING, CREATE_MONUMENT_SUCCESS, CREATE_MONUMENT_ERROR } from "../constants";
+import { LOCATION_CHANGE } from "connected-react-router";
```
njd9419, line 2: Minor: Use `'` instead of `"`

client/src/reducers/duplicates.js
```diff
@@ -1,4 +1,5 @@
 import { FETCH_DUPLICATES_PENDING, FETCH_DUPLICATES_SUCCESS, FETCH_DUPLICATES_ERROR } from '../constants';
+import { LOCATION_CHANGE } from "connected-react-router";
```
njd9419, line 2: Minor: Use `'` instead of `"`

client/src/reducers/errors.js
```diff
@@ -1,10 +1,13 @@
 import { ADD_ERROR, REMOVE_ERROR } from '../constants';
+import { LOCATION_CHANGE } from "connected-react-router";
```
njd9419, line 2: Minor: Use `'` instead of `"`

njd9419:
Code review done. Everything looks good just some minor things.

Going to check it out locally and double-check the reducers that you changed.

client/src/App.js
```diff
@@ -4,6 +4,7 @@ import './theme.scss';
 import { BrowserRouter as Router, Route } from 'react-router-dom';
```
njd9419, line 4: Minor: `Router` is not used anywhere

client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
```diff
@@ -1025,49 +1030,66 @@ export default class CreateOrUpdateForm extends React.Component {
...
-                                type="text"
-                                name="address"
-                                placeholder="Address"
-                                value={address.value}
-                                onChange={(event) => this.handleInputChange(event)}
-                                isInvalid={!address.isValid}
-                                className="text-control"
-                            />
-                            <Form.Control.Feedback type="invalid">{address.message}</Form.Control.Feedback>
+                        <span className="font-weight-bold">Location Type:</span>
```
njd9419, line 1033: Minor: Just noticed this while testing out the reducers...I think that this `<span>` can have a little less top padding/margin/whatever it has

njd9419:
Finished doing local testing of the reducers. Things look good. I didn't notice anything funky when switching pages and the issue I noticed with update is gone.

PR #115, opened 2020-04-04
Site Content Changes

njd9419:
This PR changes the site content based on the feedback we got from Dr. Decker.

There's a couple things I was unsure about so I'm going to leave them as a comment.

### DIFF
--- a/client/src/components/About/AboutInformation/AboutInformation.js
+++ b/client/src/components/About/AboutInformation/AboutInformation.js
@@ -67,19 +67,21 @@ export default class AboutInformation extends React.Component {
                     examining 26 states or territories of the US. The following fall, software engineering students in the
                     Golisano College of Computer and Information Sciences built this web application to house data and to
                     generate further interest and data collection. The spring 2020 history course added to the existing
-                    data by contributing 900 additional monuments and memorials, so that all states are represented, as
+                    data by contributing 900 additional monuments + memorials, so that all states are represented, as
                     are two territories (Guam and Puerto Rico).
                 </p>
                 {/* TODO: Replace click here with link to sign-in page */}
                 <p>
-                    Our goal is to document monuments + memorials throughout the United States and territories. We are
-                    looking for contributors to add information, please join us by creating an account (sign-in link here)
-                    or by contacting the organizers <span className="email-link">contact@monuments.us.org</span>. To
-                    view monuments and memorials, {mapNavLink}.
+                    Our goal is to document all temporary and permanent monuments + memorials throughout the United
+                    States and territories. We are looking for contributors to add information, please join us by
+                    creating an account (sign-in link here) or by contacting the organizers
+                    &nbsp;<span className="email-link">contact@monuments.us.org</span>. To view monuments and memorials,
+                    &nbsp;{mapNavLink}.
                 </p>
                 <h3 className="font-italic">
                     Background:
                 </h3>
+                {/* TODO: Add link to Vietnam Veterans Memorial */}
                 <p>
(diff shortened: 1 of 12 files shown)


client/src/components/About/AboutInformation/AboutInformation.js
```diff
@@ -27,13 +27,20 @@ export default class AboutInformation extends React.Component {
...
                 e.preventDefault();
                 window.location.replace('/map');
             }} to="/map" key="map">click here</NavLink>
         );
 
+        const nineElevenMemorialLink = (
+            <NavLink onClick={e => {
+                e.preventDefault();
+                window.location.replace('/monuments/2732');
+            }} to="/monuments/2732" key="911-memorial">9/11 Memorial</NavLink>
```
njd9419, line 41: Should we be doing the `NavLink` to a particular Monument like this? It won't work locally but it's the easy way to do it. Otherwise we'd have to search for it and thee's no guarantee the search would return what we want. Thoughts?

njd9419: I don't think anticipate the ID changing on production but who knows.

client/src/components/About/AboutInformation/AboutInformation.js
```diff
@@ -53,35 +60,45 @@ export default class AboutInformation extends React.Component {
...
+                    generate further interest and data collection. The spring 2020 history course added to the existing
+                    data by contributing 900 additional monuments + memorials, so that all states are represented, as
+                    are two territories (Guam and Puerto Rico).
+                </p>
+                {/* TODO: Replace click here with link to sign-in page */}
+                <p>
+                    Our goal is to document all temporary and permanent monuments + memorials throughout the United
+                    States and territories. We are looking for contributors to add information, please join us by
+                    creating an account (sign-in link here) or by contacting the organizers
+                    &nbsp;<span className="email-link">contact@monuments.us.org</span>. To view monuments and memorials,
```
njd9419, line 78: Is there a better way to do links to emails? I just added some styling to it to make it look like a link but you can't actually click it, which may be a bad UI.

njd9419: Actually, just noticed how @benvogler did email links on the Manage Users Page. I changed how I was doing them to do the same.

client/src/components/About/AboutInformation/AboutInformation.js
benvogler: This feels very unsafe. I think when we load all the stats we could include this in the payload by doing a search on the backend for the exact name and approximate location. If somehow there's 2 results just take the first one. If there's no results (for example locally) just don't link to anything I guess

njd9419: Sounds good. I'll make that change now.

njd9419: Done.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -684,10 +684,12 @@ private Predicate buildDecadeQuery(CriteriaBuilder builder, Root root, Integer d
 
     /**
      * Gathers the various statistics related to Monuments for the About Page
+     * @param searchForSpecificMonuments - True to also include searching for the specific Monuments we display links
+     * to on the About Page, such as the 9/11 Memorial, False otherwise
```
benvogler, line 688: You might want to add that this override mostly exists because of an H2 limitation

benvogler:
Thanks for fixing that. This feels a lot better than hard coding an id, even if it's kind of janky itself 😄

PR #119, opened 2020-04-11
Suggestions UI

njd9419:
This PR adds the UI to:

1. View Suggestions (both on the Account Page and in the Admin Panel)
2. Search Suggestions
3. Approve Suggestions
4. Reject Suggestions

It also allows Researchers and Admins to bypass the Suggestion process and directly make edits to the system.

Some of the code (especially the front-end) got a little hairy in this PR so whatever feedback you have it clean it up is appreciated. I'll point out some of the areas I think could be improved.

There are 2 current issues (I think they stem from the same root issue) that I know about:

- Clicking the link to one of the other Pending Update Suggestions in the Admin Panel requires a refresh to load data
- Clicking the link to one of the Create Suggestions inside a Bulk-Create Suggestion in the Admin Panel requires a refresh to load data

**Screenshots:**

**Warning when a user that is not email verified tries to access the Create or Update Page:**
Email Verification Warning

**Viewing Suggestions on the Account Page -- No Suggestions:**
User Page -- No Suggestions

**Viewing Suggestions on the Account Page -- Approved, Pending and Not Approved Suggestions:**
User Page -- Approved Suggestions
User Page -- Pending Suggestions
User Page -- Rejected Suggestions

**Page Displayed once a User Completes Creating a Suggestion:**
Suggestion Created Page

**Sidebar with Pending Suggestions Tooltip in Admin Panel:**
Manage Suggestions -- Panel Sidebar

**Searching Suggestions:**
Suggestion Searching

**Manage a CreateMonumentSuggestion:**
Manage CreateMonumentSuggestion

**Approved Suggestion:**
Approved Suggestion

**Rejected Suggestion:**
Rejected Suggestion

**Manage UpdateMonumentSuggestion:**
Manage Update Suggestion

**Manage BulkCreateMonumentSuggestion:**
Manage Bulk-Create Suggestion

**All Approved BulkCreateMonumentSuggestion:**
Bulk-Create All Approved

PR #121, opened 2020-04-13
Suggestions

njd9419:
This PR is a culmination of:

1. Back-End Suggestions: #112 
2. Front-End Suggestions: #113 
3. Suggestions UI #119 
4. PR feedback from #119 

I did a first round of changes from the feedback from #119 but there are still some outstanding issues. I'll add the comments back so we can discuss them.

It'd probably be useful to only review the latest commits since you've already reviewed everything else.

Also, the `LOCATION_CHANGE` issue still happens, where going from one page link to a very similar one requires a page refresh.


client/src/actions/bulk.js
```diff
@@ -67,10 +76,61 @@ function doAction(action, form, isAsyncJob) {
...
+                }, 200);
+            });
+            window.clearInterval(interval);
+
+            approveResult = await (await fetch(`${actions.create.approveUri}/approve/result/${approveJobId}`)).json();
+            dispatch(success(actions.create, approveResult));
+        } catch (err) {
+            dispatch(error(actions.create, err));
+        }
+    }
```
njd9419, line 135: image

client/src/pages/AdminPage/ManageSuggestionsPage/ManageSuggestionsPage.js
```diff
@@ -0,0 +1,114 @@
...
+            location: { search }, rejectBulkCreateSuggestion, approveBulkCreateSuggestion } = this.props;
+        const type = QueryString.parse(search).type;
+
+        const suggestion = approveCreateSuggestion.result || rejectCreateSuggestion.result ||
+            approveUpdateSuggestion.result || rejectUpdateSuggestion.result || rejectBulkCreateSuggestion.result ||
+            fetchCreateSuggestion.result || fetchUpdateSuggestion.result || fetchBulkCreateSuggestion.result;
+
+        const pending = approveCreateSuggestion.pending || rejectCreateSuggestion.pending ||
+            approveUpdateSuggestion.pending || rejectUpdateSuggestion.pending || rejectBulkCreateSuggestion.pending ||
+            fetchCreateSuggestion.pending || fetchUpdateSuggestion.pending || fetchBulkCreateSuggestion.pending;
```
njd9419, line 90: image

client/src/reducers/suggestions.js
```diff
@@ -0,0 +1,328 @@
...
+            return {
+                ...state,
+                pending: false,
+                error: action.error
+            };
+        default:
+            return state;
+    }
+}
+
```
njd9419, line 46: image

client/src/actions/bulk.js
benvogler: I guess what I meant is that there is logic in `doAction` to do singular AsyncJobs. Since `bulkCreateMonuments` does more than just that I suppose that wouldn't work. But are we even using that functionality of `doAction` anymore? You could remove it from `doAction` and instead extract some of the functionality in `bulkCreateMonuments` that's done twice (once for create, once for approve) into a utility function that `bulkCreateMonuments` calls twice.

You could also just delete the AsyncJob stuff from `doAction` and move on. I'm fine with that too.

client/src/pages/AdminPage/ManageSuggestionsPage/ManageSuggestionsPage.js
benvogler: I guess I don't fully understand what the original is doing. Is it that they are all always falsey except for one, so the defined one is always returned? in that case you could do `reducers.find(reducer => reducer.result);`

benvogler: for `pending` I do think `reducers.filter(reducer => reducer.result).length > 0` is correct - you're trying to see if at least one is pending, no?

client/src/reducers/suggestions.js
benvogler: It does that to give us the most flexibility. There are many places where we do `dispatch(success(actions.blah, {payload: result}));` instead of `dispatch(success(actions.blah, result));`, which will give you the result you're looking for I believe (where `state.blah.payload` exists in `mapStateToProps`)

client/src/actions/bulk.js
njd9419: Yeah, it's still used in `doAction`. See line 83: `bulkCreateSuggestions`. Creating the Bulk-Create Suggestions and inserting them are both `AsyncJob`s. I'm happy to make the logic generic, but the logic calls `dispatch(pending(action, progress))`. Wouldn't I need to pass `dispatch` as a parameter?

client/src/pages/AdminPage/ManageSuggestionsPage/ManageSuggestionsPage.js
njd9419: For your first comment: kind of. That's true until you approve a Suggestion. Then `approveCreateSuggestion.result` and `fetchCreateSuggestion.result` both are truthy, but we want `approveCreateSuggestion.result` since it's the updated one hence the specific ordering.
If `find` always traverses the list in order than that won't be an issue.

njd9419: Yeah, you're right for `pending`. I'll fix that

client/src/reducers/suggestions.js
njd9419: Ok, changed them all to use `basicReducer` besides the one for approve Bulk-Create Suggestions since it utilizes `progress` and `basicReducer` doesn't account for that.

client/src/utils/string-util.js
```diff
@@ -19,7 +19,7 @@ export function prettyPrintDateString(dateString) {
     if (!dateString) return;
 
     // Wednesday, October 16th, 2019 format
-    return moment(dateString).format('dddd, MMMM Do, YYYY');
+    return moment(dateString, "YYYY-MM-DD").format('dddd, MMMM Do, YYYY');
```
njd9419, line 22: I had to specify the format of the dateString otherwise Moment gave a deprecation warning. More info here: https://momentjs.com/guides/#/warnings/js-date/

client/src/actions/bulk.js
benvogler: You can pass dispatch as a param, that's actually how redux-thunk reducers work (you pass a function that accepts the param `dispatch`, and thunk calls your function passing `dispatch` in). I don't think this refactoring is necessary

client/src/pages/AdminPage/ManageSuggestionsPage/ManageSuggestionsPage.js
benvogler: Ok, that's a little complex logic there hidden by the simple `||`'s. I would leave a comment explaining that and leave them as is.

client/src/utils/string-util.js
benvogler: YYYY-MM-DD probably isn't on the list of formats it tries. Oh well.

client/src/pages/AdminPage/ManageSuggestionsPage/ManageSuggestionsPage.js
njd9419: Good point. I wasn't sure if there was a better way to do it.

client/src/utils/string-util.js
njd9419: Yeah basically it's an ISO or RSA approved format so it falls back on using a JS Date which is soon to be deprecated.

PR #122, opened 2020-04-13
Basic SEO Sitemap

benvogler:
I am trying to wade into the world of SEO slightly to make sure the site is at least indexed by Google, one of the steps in that process is creating an XML sitemap which describes the pages on your site. I created a really simple one with our main static URLs and then a dynamic list of all the monument pages. This is a little iffy because it uses the url format `https://monuments.us.org/monuments/{id}` instead of the full `https://monuments.us.org/monuments/{id}/{slug}` but I think it will still work correctly.

The sitemap looks like this (with a bunch more monuments below). In production the URLs will be `https://monuments.us.org/` not `http://localhost:3000/`

### DIFF
--- a/src/main/java/com/monumental/controllers/SitemapController.java
+++ b/src/main/java/com/monumental/controllers/SitemapController.java
@@ -36,7 +36,6 @@ public class SitemapController {
             this.publicUrl + "/",
             this.publicUrl + "/about",
             this.publicUrl + "/map",
-            this.publicUrl + "/create",
             this.publicUrl + "/search"
         );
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/monumental/controllers/SitemapController.java
```diff
@@ -0,0 +1,51 @@
...
+        response.setContentType(APPLICATION_XML_VALUE);
+
+        WebSitemapGenerator sitemap = new WebSitemapGenerator(this.publicUrl);
+
+        sitemap.addUrls(
+            this.publicUrl + "/",
+            this.publicUrl + "/about",
+            this.publicUrl + "/map",
+            this.publicUrl + "/create",
+            this.publicUrl + "/search"
```
njd9419, line 40: Any reason `/login` and `/signup` aren't here?

benvogler: It's just pages you want to have show up in search results. Now that I say that, I might get rid of `/create`

PR #123, opened 2020-04-18
Bug Fixes

njd9419:
This PR:

- Adds an extra error message for Coordinates in DMS format for Create and Bulk-Create

**Create Screenshot:**
Coordinates DMS Format Warning

**Bulk-Create Screenshot:**
Coordinates DMS Format Warning Bulk-Create

- Enforces the maximum image file size limit of 5MB for Create and Bulk-Create

**Create Screenshot:**
Create and Update File Size Too Large

**Bulk-Create Screenshot:**
Bulk-Create File Size Too Large

- Fixes the TagsSearch Component so duplicate Tags/Materials are no longer created
- Adds the most used Tag and Material as Statistic Cards to the About Page

**Screenshot (the text should be green in this picture, it's from before I made the text links):**
Most Popular Tag and Material About Page

- Fixes the Page centering on the Create and Update Pages
- Fixes potential `ArrayIndexOutOfBoundsException`s when performing geocoding and reverse geocoding
- Adds an error message when the specified Coordinates are not near the United States for Create and Bulk-Create
- Fixes broken duplicate monument links during Create

**EDIT:**
I also fixed all other usages of the `SearchResults` and `SearchResult` component so that they default to using `/monuments` as the `monumentUri` and no longer link to an undefined page (was causing the Favorites bug Dr. Decker brought up recently).

### DIFF
--- a/client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
+++ b/client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
@@ -326,6 +326,11 @@ export default class CreateOrUpdateForm extends React.Component {
                 latitude.isValid = false;
                 latitude.message = 'Latitude must not be blank';
                 formIsValid = false;
+            } else if (latitude.value.includes('°')) {
+                latitude.isValid = false;
+                latitude.message = 'Please use decimal coordinates, not degrees. ' +
+                    'To convert, input your degrees into Google Maps and copy the new numbers here.';
+                formIsValid = false;
             } else if (!validator.matches(latitude.value, latitudeRegex)) {
                 latitude.isValid = false;
                 latitude.message = 'Latitude must be valid';
@@ -335,7 +340,13 @@ export default class CreateOrUpdateForm extends React.Component {
                 longitude.isValid = false;
                 longitude.message = 'Longitude must not be blank';
                 formIsValid = false;
-            } else if (!validator.matches(longitude.value, longitudeRegex)) {
+            } else if (longitude.value.includes('°')) {
+                longitude.isValid = false;
+                longitude.message = 'Please use decimal coordinates, not degrees. ' +
+                    'To convert, input your degrees into Google Maps and copy the new numbers here.';
+                formIsValid = false;
+            }
+            else if (!validator.matches(longitude.value, longitudeRegex)) {
                 longitude.isValid = false;
                 longitude.message = 'Longitude must be valid';
                 formIsValid = false;
@@ -1107,8 +1118,9 @@ export default class CreateOrUpdateForm extends React.Component {
                         <ImageUploader
                             withIcon={false}
                             imgExtension={['.jpg', '.png']}
+                            maxFileSize={5000000}
                             label=""
-                            fileSizeError="File size is too large"
+                            fileSizeError="File size is too large. Maximum file size is 5MB"
                             fileTypeError="File type is not supported"
                             withPreview={true}
                             onChange={(files) => this.handleImageUploaderChange(files)}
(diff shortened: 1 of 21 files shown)


client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
```diff
@@ -326,19 +326,49 @@ export default class CreateOrUpdateForm extends React.Component {
...
+                formIsValid = false;
             } else if (!validator.matches(latitude.value, latitudeRegex)) {
                 latitude.isValid = false;
                 latitude.message = 'Latitude must be valid';
                 formIsValid = false;
+            } else {
+                const latAsDouble = parseFloat(latitude.value);
+                // Alaska is the furthest north location and its latitude is approximately 71
+                // The American Samoa is the furthest south location and its latitude is approximately -14
+                if (latAsDouble > 72 || latAsDouble < -15) {
```
njd9419, line 342: @benvogler I'd appreciate you double-checking these numbers and the logic

client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
```diff
@@ -326,19 +326,49 @@ export default class CreateOrUpdateForm extends React.Component {
...
+            }
+            else if (!validator.matches(longitude.value, longitudeRegex)) {
                 longitude.isValid = false;
                 longitude.message = 'Longitude must be valid';
                 formIsValid = false;
+            } else {
+                const lonAsDouble = parseFloat(longitude.value);
+                // Guam is the furthest west location and its longitude is approximately 144
+                // Puerto Rico is the furthest east location and its longitude is approximately -65
+                if (lonAsDouble > -64 && !(lonAsDouble < 180 && lonAsDouble > 143)) {
```
njd9419, line 367: @benvogler Same here

client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
benvogler: This is the page I mentioned last week - https://en.wikipedia.org/wiki/List_of_extreme_points_of_the_United_States

Based on that, this looks good

client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
benvogler: Damn prime meridian making this boolean logic all confusing. I think this is right

client/src/components/Search/TagsSearch/TagsSearch.js
```diff
@@ -61,9 +61,10 @@ class TagsSearch extends React.Component {
           <div/>
         );
 
-        if (allowTagCreation && searchQuery !== '' && !selectedTags.find(tag => tag.name === searchQuery)
-            && !createdTags.find(tag => tag.name === searchQuery))
-        {
+        if (allowTagCreation && searchQuery !== ''
+            && !selectedTags.find(tag => tag.name.toLowerCase() === searchQuery.toLowerCase())
+            && !createdTags.find(tag => tag.name.toLowerCase() === searchQuery.toLowerCase())
+            && !visibleSearchResults.find(tag => tag.name.toLowerCase() === searchQuery.toLowerCase())) {
```
benvogler, line 67: This is good. But, say I input `stone` and there's no `Stone` or other casing. Would it be added as `stone` or do we auto-capitalize to `Stone`? Because I think we should capitalize

benvogler: Might also want to double check how this is handled for bulk

PR #130, opened 2020-04-27
Bug Fixes Round 4

benvogler:
* Better handling of US territories for the State field
    * Google Maps considers Guam and Puerto Rico countries, so they don't show up in the same places states do. We now have a somewhat intelligent fallback mechanism where if the country has a 2 letter short code and the state didn't exist or wasn't 2 letters we use the country as the state
* Fixed search result numbers not changing after a filter change reset the page to 1 (you could have result 101 at the top of page 1 because of this)
* Show no results instead of old results if the LocationSearch got no results back
    * I wanted to show a "No results found" line but it doesn't seem possible with the package we're using
* Fix images overflowing the modal when you click to expand them on the monument page
* Fix a bunch of react warnings
* Fix monument page not querying for new monument after changing from one monument to another (related/nearby monuments, etc.)
* Fix the clear button in text/location search overlapping the text in the input
* Fix url not updating with monument slug on monument page
* Replace usages of `history.replace` with `history.push` so that you can go back in your browser
* When you try to access a protected route without being logged in, it sends to you the login page and will redirect you back after you login. For pages with path params this didn't work correctly, but it does now
* Fix null descriptions making score null in searches
* Add `word_similarity` to relevance search and make title 3x as important as artist or description
* Safely handle larger than acceptable latitude/longitudes in bulk upload (a value of over 90,000 was causing a fatal error)
* Improve latitude/longitude error handling so that you don't get cascading errors/warnings because of one fault. For example if you used degrees, you would get a warning about that and then an error because there was no value saved. Now, the error is suppressed and that warning is elevated to an error so that you still can't continue, but you aren't as confused by the messages
* Changed the copyright in the footer to automatically be the current year
* Allow "Do Not Map" to be selected multiple times in bulk upload mapping, just like "Select a field"
* Auto remove blank columns in bulk upload field mapping
* Redirect HTTP requests to HTTPS in production. This fixes that mysterious bug where typing monuments.us.org into your browser would get you nowhere
* Get city/state when geocoding/reverse geocoding on the backend, like we do already on the frontend

### DIFF
--- a/client/src/components/AdminPanel/ManageSuggestions/SuggestionSearch/SuggestionSearch.js
+++ b/client/src/components/AdminPanel/ManageSuggestions/SuggestionSearch/SuggestionSearch.js
@@ -10,10 +10,12 @@ export default class SuggestionSearch extends React.Component {
         const { showSearchResults, onLimitChange, onPageChange, limit, page, count, suggestions } = this.props;
         const pageCount = Math.ceil(count / limit);
 
+        console.log(page);
+
         return (
             <div className="suggestion-search">
                 <div className="sticky-top">
-                    <SuggestionSearchBar/>
+                    <SuggestionSearchBar page={page} limit={limit}/>
                     {showSearchResults &&
                         <div className="mt-2">
                             <SearchInfo onLimitChange={onLimitChange} limit={limit} page={page} count={count} hideSortBy/>
--- a/client/src/components/AdminPanel/ManageSuggestions/SuggestionSearch/SuggestionSearch.js
+++ b/client/src/components/AdminPanel/ManageSuggestions/SuggestionSearch/SuggestionSearch.js
@@ -10,8 +10,6 @@ export default class SuggestionSearch extends React.Component {
         const { showSearchResults, onLimitChange, onPageChange, limit, page, count, suggestions } = this.props;
         const pageCount = Math.ceil(count / limit);
 
-        console.log(page);
-
         return (
             <div className="suggestion-search">
                 <div className="sticky-top">
--- a/client/src/components/BulkCreateForm/BulkCreateForm.js
+++ b/client/src/components/BulkCreateForm/BulkCreateForm.js
@@ -186,7 +186,6 @@ export default class BulkCreateForm extends React.Component {
                         mappedField
                     }
                 }).filter(pair => pair);
-                console.log('made mapping', mapping);
                 this.setState({mapping, fields})
             });
     }
--- a/client/src/components/BulkCreateForm/BulkCreateForm.js
+++ b/client/src/components/BulkCreateForm/BulkCreateForm.js
@@ -164,6 +164,7 @@ export default class BulkCreateForm extends React.Component {
                     await this.setState({showFieldMapping: true});
                 }
                 const mapping = headers.map(header => {
+                    if (!header) return;
                     let mappedField = '';
                     for (let field of fields) {
                         // By default don't select images on CSV uploads since they don't work.
@@ -184,8 +185,8 @@ export default class BulkCreateForm extends React.Component {
                         originalField: header,
                         mappedField
                     }
-                });
-
+                }).filter(pair => pair);
+                console.log('made mapping', mapping);
                 this.setState({mapping, fields})
             });
     }
@@ -360,11 +361,13 @@ export default class BulkCreateForm extends React.Component {
                             <div className="column d-flex justify-content-end my-1">
                                 <Form.Control as="select" className="mr-3" value={pair.mappedField} onChange={event => {
                                     pair.mappedField = event.currentTarget.value;
-                                    mapping.forEach((currentPair, currentIndex) => {
-                                        if (currentPair.mappedField === pair.mappedField && index !== currentIndex) {
-                                            currentPair.mappedField = '';
-                                        }
-                                    });
+                                    if (pair.mappedField !== 'Do Not Map') {
+                                        mapping.forEach((currentPair, currentIndex) => {
+                                            if (currentPair.mappedField === pair.mappedField && index !== currentIndex) {
+                                                currentPair.mappedField = '';
+                                            }
+                                        });
+                                    }
                                     fields.forEach(field => {
                                         field.selected = !!mapping.find(currentPair => {
                                             return currentPair.mappedField === field.name;
@@ -491,7 +494,7 @@ export default class BulkCreateForm extends React.Component {
                         will <span className="font-weight-bold">not</span> be {pastTenseTerm.toLowerCase()}.
                         {warningCount > 0 &&
                             <span>
-                                Any rows with warnings will still be {pastTenseTerm.toLowerCase()}, but may have
+                                &nbsp;Any rows with warnings will still be {pastTenseTerm.toLowerCase()}, but may have
                                 non-critical issues that should be addressed with updates later.
                             </span>
                         }
--- a/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
+++ b/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
@@ -16,7 +16,7 @@ export default class Map extends React.Component {
             q = [monument.lat, monument.lon].join(',');
         }
 
-        q = escape(q);
+        q = encodeURIComponent(q);
 
         return (
             <iframe title="gmaps-iframe"
--- a/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
+++ b/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
@@ -9,7 +9,10 @@ export default class Map extends React.Component {
         const monument = this.props.monument;
 
         let q = monument.address;
-        if (monument.address && monument.address.includes('Unnamed Road') || (!q && monument.lat && monument.lon)) {
+        if ((monument.address &&
+            (monument.address.includes('Unnamed Road') ||
+            monument.address.startsWith([monument.city, monument.state].join(', ')))) ||
+            (!q && monument.lat && monument.lon)) {
             q = [monument.lat, monument.lon].join(',');
         }
 
--- a/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
+++ b/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
@@ -13,6 +13,8 @@ export default class Map extends React.Component {
             q = [monument.lat, monument.lon].join(',');
         }
 
+        q = escape(q);
+
         return (
             <iframe title="gmaps-iframe"
                     src={`https://maps.google.com/maps?q=${q}&z=16&output=embed`}
--- a/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
+++ b/client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
@@ -9,7 +9,9 @@ export default class Map extends React.Component {
         const monument = this.props.monument;
 
         let q = monument.address;
-        if (!q && monument.lat && monument.lon) q = [monument.lat, monument.lon].join(',');
+        if (monument.address && monument.address.includes('Unnamed Road') || (!q && monument.lat && monument.lon)) {
+            q = [monument.lat, monument.lon].join(',');
+        }
 
         return (
             <iframe title="gmaps-iframe"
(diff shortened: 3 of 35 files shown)


client/src/components/AdminPanel/ManageSuggestions/SuggestionSearch/SuggestionSearch.js
```diff
@@ -10,10 +10,12 @@ export default class SuggestionSearch extends React.Component {
         const { showSearchResults, onLimitChange, onPageChange, limit, page, count, suggestions } = this.props;
         const pageCount = Math.ceil(count / limit);
 
+        console.log(page);
```
njd9419, line 13: Debug print

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -184,8 +185,8 @@ export default class BulkCreateForm extends React.Component {
                         originalField: header,
                         mappedField
                     }
-                });
-
+                }).filter(pair => pair);
+                console.log('made mapping', mapping);
```
njd9419, line 189: Debug print

client/src/components/Suggestions/BulkCreateMonumentSuggestions/BulkCreateMonumentSuggestion/BulkCreateMonumentSuggestion.js
```diff
@@ -1,38 +1,52 @@
...
  * Presentational component for displaying a BulkCreateMonumentSuggestion
  */
-export default class BulkCreateMonumentSuggestion extends React.Component {
+class BulkCreateMonumentSuggestion extends React.Component {
+
+    static mapStateToProps(state) {
+        return {
+            session: state.session
+        };
+    }
```
njd9419, line 19: Do you think the user role should be passed down rather than having this component aware of Redux?

client/src/components/Suggestions/CreateMonumentSuggestions/CreateMonumentSuggestion/CreateMonumentSuggestion.js
```diff
@@ -19,6 +21,12 @@ export default class CreateMonumentSuggestion extends React.Component {
         };
     }
 
+    static mapStateToProps(state) {
+        return {
+            session: state.session
+        };
+    }
```
njd9419, line 28: Again, should this be passed down via props?

client/src/components/Suggestions/UpdateMonumentSuggestions/UpdateMonumentSuggestion/UpdateMonumentSuggestion.js
```diff
@@ -1,11 +1,19 @@
...
+import { Role } from '../../../../utils/authentication-util';
 
-export default class UpdateMonumentSuggestion extends React.Component {
+class UpdateMonumentSuggestion extends React.Component {
+
+    static mapStateToProps(state) {
+        return {
+            session: state.session
+        };
+    }
```
njd9419, line 16: Same thing, maybe pass down as prop instead

client/src/pages/AdminPage/ManageSuggestionsPage/SuggestionSearchPage/SuggestionSearchPage.js
```diff
@@ -73,7 +73,9 @@ class SuggestionSearchPage extends React.Component {
             this.setState({type: params.type});
         }
         if (prevProps.location.search !== search) {
-            dispatch(searchSuggestions(QueryString.parse(search)));
+            const queryParams = QueryString.parse(search);
+            this.setState({...queryParams});
```
njd9419, line 77: Should this have an `await`?

client/src/pages/AdminPage/ManageSuggestionsPage/SuggestionSearchPage/SuggestionSearchPage.js
```diff
@@ -94,7 +96,7 @@ class SuggestionSearchPage extends React.Component {
 
     async search(changedState) {
         await this.setState({changedState});
-        search(this.state, this.props.history, '/panel/manage/suggestions/search');
+        search({...this.state, ...changedState}, this.props.history, '/panel/manage/suggestions/search');
```
njd9419, line 99: What's the point of doing this if there's the `await this.setState({changedState});` line above?

src/main/java/com/monumental/security/ProductionHTTPRedirect.java
```diff
@@ -0,0 +1,42 @@
...
+            protected void postProcessContext(Context context) {
+                SecurityConstraint securityConstraint = new SecurityConstraint();
+                securityConstraint.setUserConstraint("CONFIDENTIAL");
+                SecurityCollection collection = new SecurityCollection();
+                collection.addPattern("/*");
+                securityConstraint.addCollection(collection);
+                context.addConstraint(securityConstraint);
+            }
+        };
+        tomcat.addAdditionalTomcatConnectors(redirectConnector());
```
njd9419, line 30: Minor: Could use `this.redirectConnector()`

src/main/java/com/monumental/services/GoogleMapsService.java
```diff
@@ -41,7 +46,7 @@ public String getAddressFromCoordinates(Double lat, Double lon) {
      * @param address - String for the address to geocode
      * @return Geometry - Google Maps Geometry containing the location data for the address
```
njd9419, line 47: Update the docstring here. We should probably add a docstring for `getAddressFromCoordinates` too.

src/main/java/com/monumental/services/GoogleMapsService.java
```diff
@@ -64,6 +69,44 @@ public Geometry getCoordinatesFromAddress(String address) {
...
+           For Guam, there is no administrative_area_level_1, and Guam is the country
+           This is kind of a catch all attempt to get a 2 letter state code
+         */
+        if ((bundle.state == null || bundle.state.length() != 2) && country != null && country.length() == 2) {
+            bundle.state = country;
+        }
+        /* If there was no 2 letter state code, we could end up with something longer like San Juan, so fallback to
+           leaving the field blank if it's not 2 letters long
+         */
+        if (bundle.state == null || bundle.state.length() != 2) {
```
njd9419, line 101: I think the first part of this `if` is redundant

njd9419: It should probably be `bundle.state != null && bundle.state.length() != 2`

client/src/components/Monument/MapPhotoSphereTabs/Map/Map.js
```diff
@@ -9,7 +9,14 @@ export default class Map extends React.Component {
...
         let q = monument.address;
-        if (!q && monument.lat && monument.lon) q = [monument.lat, monument.lon].join(',');
+        if ((monument.address &&
+            (monument.address.includes('Unnamed Road') ||
+            monument.address.startsWith([monument.city, monument.state].join(', ')))) ||
+            (!q && monument.lat && monument.lon)) {
+            q = [monument.lat, monument.lon].join(',');
+        }
+
+        q = escape(q);
```
njd9419, line 19: According to a quick Google search, `escape()` is deprecated.

njd9419: https://www.w3schools.com/jsref/jsref_escape.asp

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -1379,16 +1379,20 @@ public void updateMonumentTags(Monument monument, List<String> newTagNames, bool
     public void populateNewMonumentLocation(Monument monument) {
         // If the Monument has no address, do a reverse geocode
         if (monument.getAddress() == null && monument.getCoordinates() != null) {
-            String address = this.googleMapsService.getAddressFromCoordinates(monument.getLat(), monument.getLon());
-            if (address != null) {
-                monument.setAddress(address);
+            GoogleMapsService.AddressBundle bundle = googleMapsService.getAddressFromCoordinates(monument.getLat(), monument.getLon());
```
njd9419, line 1382: Minor: `this.googleMapsService`

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -1450,9 +1454,11 @@ private void populateUpdatedMonumentAddress(Monument newMonument, String oldAddr
         }
 
         // Perform reverse geocoding
-        String address = this.googleMapsService.getAddressFromCoordinates(newMonument.getLat(), newMonument.getLon());
-        if (address != null) {
-            newMonument.setAddress(address);
+        GoogleMapsService.AddressBundle bundle = googleMapsService.getAddressFromCoordinates(newMonument.getLat(), newMonument.getLon());
```
njd9419, line 1457: Minor: `this.googleMapsService`

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -80,17 +85,21 @@
...
-                            // Alaska is the furthest north location and its latitude is approximately 71
-                            // The American Samoa is the furthest south location and its latitude is approximately -14
-                            if (latitude > 72 || latitude < -15) {
-                                result.getErrors().add("Latitude is not near the United States");
+                                // Alaska is the furthest north location and its latitude is approximately 71
+                                // The American Samoa is the furthest south location and its latitude is approximately -14
+                                if (latitude > 72 || latitude < -15) {
+                                    result.getErrors().add("Latitude is not near the United States");
+                                }
+                                // If the not a valid latitude, set it to null because extremely large values can break everything
```
njd9419, line 96: Minor: Fix grammar of comment

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -101,17 +110,21 @@
...
-                            // Guam is the furthest west location and its longitude is approximately 144
-                            // Puerto Rico is the furthest east location and its longitude is approximately -65
-                            if (longitude > -64 && !(longitude < 180 && longitude > 143)) {
-                                result.getErrors().add("Longitude is not near the United States");
+                                // Guam is the furthest west location and its longitude is approximately 144
+                                // Puerto Rico is the furthest east location and its longitude is approximately -65
+                                if (longitude > -64 && !(longitude < 180 && longitude > 143)) {
+                                    result.getErrors().add("Longitude is not near the United States");
+                                }
+                                // If the not a valid longitude, set it to null because extremely large values can break everything
```
njd9419, line 121: Minor: Fix comment grammar

src/test/java/com/monumental/services/integrationtest/MonumentServiceMockIntegrationTests.java
```diff
@@ -186,7 +186,7 @@ public void testMonumentService_BulkCreateFlow_OneInvalidCsvRecord() {
 
         CsvMonumentConverterResult validationErrors = validationResult.getInvalidResults().get(1);
 
-        assertEquals(3, validationErrors.getWarnings().size());
+        assertEquals(1, validationErrors.getWarnings().size());
         assertTrue(validationErrors.getWarnings().contains("Please use decimal coordinates, not degrees. To convert, input your degrees into Google Maps."));
```
njd9419, line 190: We could import the string from `CsvMonumentConverter` here and use that for `.contains()`

client/src/components/Suggestions/BulkCreateMonumentSuggestions/BulkCreateMonumentSuggestion/BulkCreateMonumentSuggestion.js
benvogler: These get rendered in a couple different places so I'd have to pass it through a bunch of components that don't actually care about the session at all. In this case, I think it's better just to get the session where we need it (and also less time consuming to write)

client/src/pages/AdminPage/ManageSuggestionsPage/SuggestionSearchPage/SuggestionSearchPage.js
benvogler: Honestly there shouldn't be a difference but this fixed the bug somehow so I'm just going to leave it

src/main/java/com/monumental/services/GoogleMapsService.java
benvogler: It's checking if we didn't find a state, or if we did and it's not two letters long, which can happen for places like San Juan, Puerto Rico where the short_name of the state is actually `San Juan`

njd9419: I know, but we're checking if `bundle.state` is `null` and if it is, we set `bundle.state` equal to `null`. Am I missing something?

benvogler: You can't check `bundle.state.length` without checking if `bundle.state` exists, otherwise you'll get an exception. I suppose `||` or `&&` both work here depending on `==` or `!=`

njd9419: I think `&&` makes more sense when you read it

PR #136, opened 2020-09-22
create pdf export

renchauret:


client/src/components/Export/ExportButtons/ExportButtons.js
```diff
@@ -0,0 +1,29 @@
...
+
+    render() {
+        const { fields, data, title } = this.props;
+
+        return (
+            <span>
+                <span>
+                    <ExportToCsvButton className="mt-2" fields={fields}
+                                       data={data}
+                                       exportTitle={`${title} Data ${moment().format('YYYY-MM-DD hh:mm')}`}/>
```
dwang305, line 19: Files replace the ":" with "_". This makes time part of the file name is sort of confusing to look at. I didn't know what it was supposed to be until I checked the code. Maybe remove it, I don't think the exact time is needed.

client/src/utils/export-util.js
```diff
@@ -7,4 +9,30 @@ import { parse as toCSV } from 'json2csv';
...
+
+/**
+ * Export the specified data with the specified fields to CSV format
+ * @param fields - Array of the names of the columns for the CSV
+ * @param data - Array of data representing the rows of the CSV
+ */
+export function exportToPdf(fields, data, exportTitle) {
+    const dataArr = []
+    for (const field in data[0]) {
+        dataArr.push(data[0][field])
```
dwang305, line 22: As mentioned before with Dr. Decker, width of columns causing weird text wrapping issues. I believe this is because of the tags column. If there is more than one tag, it will separate by commas but no space after the comma. I think the lack of space between the comma and next tag causes the spacing issues.

client/src/components/Export/ExportButtons/ExportButtons.js
renchauret: I agree, but let's ask Dr. Decker what she thinks.  This is the format that has been in use from the previous team, so she might like it

client/src/utils/export-util.js
renchauret: good catch, I will look into fixing this

renchauret: added spaces in most recent commit

dwang305 (approved):

PR #138, opened 2020-10-07
Deactivated date

renchauret:
add deactivated date to monuments
- add to monument in backend
- show on monument page
- allow adding and editing in create and update monument forms
- validate in forms that deactivated date is after created date
- add to bulk upload (not working, created date not working right now either)


dwang305 (changes requested):
For adding a deactivated date with the Month/Year option, the dropdown menu for month is not clickable. As in, the drop down appears, but none of the months can be selected. Stays as January.

dwang305 (changes requested):
Year and Month/Year don't seem to work correctly. With either option, the new date shown will have a month, day, and year.

dwang305 (commented):
No way to remove a deactivated date once it has been added to a monument. Is there any way to change that or is this expected behavior?

renchauret:
> Year and Month/Year don't seem to work correctly. With either option, the new date shown will have a month, day, and year.

This is expected and is also how created date works.  On the backend, it is still represented as a full date, just defaults to first of the month and first month of the year.  I will make a note to ask Dr. Decker what she thinks about this, maybe we should only allow full date entry

renchauret:
> No way to remove a deactivated date once it has been added to a monument. Is there any way to change that or is this expected behavior?

This is also expected behavior, as it is how date works.  Again, I can ask Dr. Decker how she feels about this.

renchauret:
> For adding a deactivated date with the Month/Year option, the dropdown menu for month is not clickable. As in, the drop down appears, but none of the months can be selected. Stays as January.

Yeah this is a bug, good catch.  Will let you know once I've fixed it

dwang305 (changes requested):
Deactivated date has a new bug where the displayed deactivated date is the day before the date specified by the user input.

dwang305 (approved):

PR #141, opened 2020-10-15
Export additions

renchauret:


dwang305 (commented):
Very minor issue, for monuments with multiple contributors, there should be a space after the comma separating them. Right now it's like Person 1,Person 2 instead of Person 1, Person 2.

dwang305 (approved):

PR #148, opened 2020-10-30
different color markers for temporary monuments

renchauret:
went with green for now


dwang305 (commented):
Potential issue: When going to the edit monuments page from manage monuments, the "Is Temporary" box is by default set to "No" even for monuments that are temporary. If you were to click the submit button, without having touched anything, the system will try to make it a permanent monument.

renchauret:
> Potential issue: When going to the edit monuments page from manage monuments, the "Is Temporary" box is by default set to "No" even for monuments that are temporary. If you were to click the submit button, without having touched anything, the system will try to make it a permanent monument.

Good call out.  This sounds like something that was already present before I added this code, so I think we should make a separate bug ticket for it.  Is this good otherwise?

dwang305:
> Good call out. This sounds like something that was already present before I added this code, so I think we should make a separate bug ticket for it. Is this good otherwise?

Yes

dwang305 (approved):

PR #151, opened 2020-11-11
Bulk Upload Work

renchauret:
1. Fix issue where only 1 of multiple errors are returned.
2. Clarify to user how to insert arrays in CSV.
3. Fix issue with bulk upload image naming causing "0 monuments" issue.
    i. Note:  This may not be only cause of "0 monuments".  I believe it occurs whenever backend does not explicitly handle an error.
4. Add CSV and Zip templates to help users.


dwang305 (commented):
For multiple errors in a zip, only 1 error is being reported to user still. Tested this out by using the template zip and added a second csv and random pdf. Multiple csv error was reported but not unsupported file type error.

renchauret:
> For multiple errors in a zip, only 1 error is being reported to user still. Tested this out by using the template zip and added a second csv and random pdf. Multiple csv error was reported but not unsupported file type error.

Interesting, good catch.  I am no longer seeing it with multiple errors within the CSV at least.  Will look into this

renchauret:
> For multiple errors in a zip, only 1 error is being reported to user still. Tested this out by using the template zip and added a second csv and random pdf. Multiple csv error was reported but not unsupported file type error.

Ok, this should be fixed now

renchauret:
This one is being really finnicky.  Nearby and related should now work after clicking the back button twice.  I think that is good enough for now.  I will create a bug ticket to improve this later

dwang305 (commented):
When there is more than one csv in the zip, everything below the "Bulk Create Monuments and Memorials" header disappears. Using inspect element, it appears that the text box that should contain the upload and error message was never generated.

dwang305 (approved):
When there is more than one csv in the zip, everything below the "Bulk Create Monuments and Memorials" header disappears. Using inspect element, it appears that the text box that should contain the upload and error message was never generated.

PR #152, opened 2020-12-08
Text Change:  'Suggest' is now 'Create' or 'Make' for Researchers and Above

renchauret:
Less than Researcher or not logged in:

Greater than or equal to Researcher:

### DIFF
--- a/client/src/components/SuggestChanges/SuggestChanges.js
+++ b/client/src/components/SuggestChanges/SuggestChanges.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Button, Card } from 'react-bootstrap';
+import { Role } from '../../utils/authentication-util';
 
 /**
  * Prompts users to suggest changes to a Monument on its record page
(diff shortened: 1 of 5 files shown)


client/src/components/User/Suggestions/Suggestions.js
```diff
@@ -103,7 +105,7 @@ export default class Suggestions extends React.Component {
             {role.toUpperCase() === Role.PARTNER && <>
                 <h6 className="mt-4">Bulk New Monument or Memorial Suggestions</h6>
                 {(!suggestions.bulk || suggestions.bulk.length === 0) && !error && <>
-                    You don't have any {type} bulk new monument or memorial suggestions yet. You can suggest bulk
+                    You don't have any {type} bulk new monument or memorial suggestions yet. You can {createText} bulk
```
wrs1492, line 108: Since this can only be viewed by partners right now, not a necessary change, but doesn't really matter.

client/src/components/SuggestChanges/SuggestChanges.js
```diff
@@ -14,25 +15,35 @@ export default class SuggestChanges extends React.Component {
...
+                                If your change is approved we'll send you an email to let you know, and you will be
+                                listed as a contributor on this page!</p>
+                        }
+                        <Button variant="primary" onClick={() => this.handleButtonClick()}>{createText} A Change</Button>
                     </>}
                     {!isUpdate && <>
                         <p>Contributions help us to expand our database of Monuments and Memorials.</p>
-                        <p>Your suggestion will be reviewed by our team and then published for the public to see. Your contribution will be credited on the monument or memorial's page.</p>
-                        <Button variant="primary" onClick={() => this.handleButtonClick()}>Suggest New</Button>
+                        <p>Your suggestion will be {!researcherOrAbove && 'viewed by our team and then '} published for the public to see. Your contribution will be credited on the monument or memorial's page.</p>
```
wrs1492, line 45: If we are being perfect here this suggestion should be also be changed to Your "change", but that would require a bit too much work to be worth it imo

wrs1492 (approved):
Are some small things that could be improved with time, but nothing worthwhile.

client/src/components/SuggestChanges/SuggestChanges.js
renchauret: yeah I'm ok with this slight misnomer as well

PR #154, opened 2020-12-10
reformat csv export to match bulk upload

renchauret:
now, when you csv export, the csv should work in bulk upload without any changes to the file

old format:

new format:


wrs1492 (commented):
Works for me, there are some linting errors that are fixed with the export-import-zip branch, so merge that in first.

wrs1492 (approved):

PR #155, opened 2020-12-16
Export import zip

renchauret:
new bulk export option:  export to zip.  It uses the improved formatting from branch `export-import` so that it works for bulk upload.  The downloaded zip folder also includes images and links them in the csv.

### DIFF
--- a/client/src/components/Export/ExportToZipButton/ExportToZipButton.js
+++ b/client/src/components/Export/ExportToZipButton/ExportToZipButton.js
@@ -19,22 +19,23 @@ export default class ExportToZipButton extends React.Component {
 
         const zip = new JSZip();
         zip.file(exportFileName, csv);
-
-        for (const image of images) {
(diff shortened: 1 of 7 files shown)


client/src/components/Export/ExportToZipButton/ExportToZipButton.js
```diff
@@ -0,0 +1,56 @@
...
+    async handleClick() {
+        const { fields, data, exportTitle, images } = this.props;
+
+        const csv = toCSV(data, {fields});
+        const exportFileName = exportTitle.endsWith('.csv') ? exportTitle : exportTitle + '.csv';
+
+        const zip = new JSZip();
+        zip.file(exportFileName, csv);
+
+        for (const image of images) {
```
wrs1492, line 23: If the no memorial in the query has an image, this errors. add a check to make sure images is not null.

wrs1492 (approved):

PR #164, opened 2021-02-11
Coordinate conversion

bcail98:


client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
```diff
@@ -418,17 +433,31 @@ export default class CreateOrUpdateForm extends React.Component {
...
+                    const lonSec = parseFloat(lonValues[2]);
+
+                    //decimal = degrees + (minutes/60) + (seconds/3600)
+                    let lonDecimal = Math.abs(lonDegree) + (lonMin/60) + (lonSec/3600);
+                    if ((longitude.value.includes('W'))||(longitude.value.includes('w'))||(lonDegree<0)) {
+                        lonDecimal *= -1;
+                    }
+                    longitude.value = lonDecimal.toString();
+                }
+            }
```
renchauret, line 455: could be a function with lat or long passed in to avoid dupe code

bcail98: Done

renchauret:
your change for my last comment looks good

played around with it a bit, couple things:
- could use some rounding, maybe to 6 decimal places. I got this off a conversion:  Coordinates: 43.08466666666667, -112.505916
- using anything other than decimal format does not trigger our address autofiller using Google Maps API.  So if you enter the same coordinates for two monuments, but in two different formats, one monument will get an address and the other will not.  Idk how complicated this is to fix.  If you think it's not too bad, add it to this, otherwise, please create a ticket for it on Trello

bcail98:
> your change for my last comment looks good
> 
> played around with it a bit, couple things:
> 
> * could use some rounding, maybe to 6 decimal places. I got this off a conversion:  Coordinates: 43.08466666666667, -112.505916
> * using anything other than decimal format does not trigger our address autofiller using Google Maps API.  So if you enter the same coordinates for two monuments, but in two different formats, one monument will get an address and the other will not.  Idk how complicated this is to fix.  If you think it's not too bad, add it to this, otherwise, please create a ticket for it on Trello

Fixed both issues.

renchauret (approved):

PR #171, opened 2021-02-26
fix export zip for chrome

renchauret:


client/src/components/Export/ExportToZipButton/ExportToZipButton.js
```diff
@@ -24,20 +24,30 @@ export const ExportToZipButton = (props) => {
...
+                        Bucket: 'monuments-and-memorials',
+                        Key: key
+                    }).promise();
+                    let name = getS3ImageNameFromObjectUrl(image.url)
+                    if (!name.endsWith('.png') && !name.endsWith('.jpg')) {
+                        name = name + '.jpg'
                     }
-                )
+                    zip.file(name, data.Body, {binary: true});
+                    console.log(data)
```
wrs1492, line 46: Does this need to be here?

client/src/components/Export/ExportToZipButton/ExportToZipButton.js
```diff
@@ -24,20 +24,30 @@ export const ExportToZipButton = (props) => {
...
+                } catch (e) {
```
wrs1492, line 47: Do we want to do anything with the caught exception?

client/src/components/Export/ExportToZipButton/ExportToZipButton.js
renchauret: removed it

client/src/components/Export/ExportToZipButton/ExportToZipButton.js
renchauret: added rollbar logging

wrs1492 (approved):

PR #172, opened 2021-03-02
Country Filter

bcail98:
Fixed and improved country filter by using Google Maps geocode data. Updated advisors on about page.


wrs1492 (changes requested):
need to move the validation logic to getAddressCityStateFromGeocodingResult so users will know before the press validate form. also a check should be added to the address select function.

wrs1492 (approved):

PR #174, opened 2021-03-07
Add Image Reference URLs and Image Captions

renchauret:
adds reference URLs and captions to images.  Also adds ability to add or edit them in single monument create/update and bulk upload


src/main/java/com/monumental/util/csvparsing/CsvMonumentConverterResult.java
```diff
@@ -176,5 +196,19 @@ public void validate() {
...
+
+        /* Image Reference URL Validation */
+        /* Check that the image references are valid URLs */
+        if (this.imageReferenceUrls != null) {
+            for (String imageReferenceUrl : this.imageReferenceUrls) {
+                try {
+                    URL url = new URL(imageReferenceUrl);
+                } catch (MalformedURLException e) {
+                    if (!this.getErrors().contains("All Image References must be valid URLs")) {
+                        this.getErrors().add("All Image References must be valid URLs");
```
wrs1492, line 208: maybe add the string to the error message of the bad url?

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
```diff
@@ -232,6 +232,20 @@
                                 result.getWarnings().add("Cannot upload images with a .csv file. You must package your .csv and your images into a .zip file and upload it.");
                             }
                             break;
+                        case "imageReferenceUrls":
+                            if (zipFile != null) {
+                                result.getImageReferenceUrls().addAll(parseCsvArray(value, true));
+                            } else {
+                                result.getWarnings().add("Cannot upload images with a .csv file. You must package your .csv and your images into a .zip file and upload it.");
```
wrs1492, line 239: Won't this just create duplicate warnings? not sure I see the point of checking for a zip file here.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -1187,7 +1197,7 @@ public Monument updateMonument(UpdateMonumentSuggestion updateSuggestion) {
      * @param arePhotoSphereImages - True if the specified imageUrls are for PhotoSphere images, False otherwise
      * @return List<Image> - List of new Images with the specified imageUrls and associated with the specified Monument
      */
-    public List<Image> createMonumentImages(List<String> imageUrls, Monument monument, boolean arePhotoSphereImages) {
+    public List<Image> createMonumentImages(List<String> imageUrls, List<String> imageReferenceUrls, List<String> imageCaptions, Monument monument, boolean arePhotoSphereImages) {
```
wrs1492, line 1200: Add new parameters to comment

wrs1492 (commented):
Haven't been able to test the export and stuff since its broken for me, let me know when thats fixed. Everything else looks good, most of my comments aren't super important to address, up to you on what you want to fix

client/src/components/Monument/Images/Gallery/Gallery.js
```diff
@@ -105,16 +105,29 @@ export default class Gallery extends React.Component {
         }
         const selectedImage = images[selectedImageIndex];
         const ariaLabel = this.getAltText(selectedImage)
+        let referenceUrl = selectedImage.referenceUrl
+        if (referenceUrl && !referenceUrl.startsWith('https://') && !referenceUrl.startsWith('http://')) {
+            referenceUrl = `https://${referenceUrl}`
```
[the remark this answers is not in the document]
renchauret: that's a good question.  I'm not sure a better way to do this though, can't really tell if the site is not secure here.  Most sites are secure at this point, and hopefully most people will be copy-pasting URLs so they include this.  Problem is that URLs like www.google.com are valid URLs, but when used in a link like this without https, the browser treats them as a relative link to page /www.google.com on our site, which obviously doesn't work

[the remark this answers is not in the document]
renchauret: there probably is some better way to do this, but will take some research so maybe worth its own ticket

client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
```diff
@@ -782,6 +874,55 @@ export default class CreateOrUpdateForm extends React.Component {
         return updateForm;
     }
 
+    validateUrl(url) {
```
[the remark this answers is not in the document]
renchauret: yeah i'll move it, and can document

src/main/java/com/monumental/util/csvparsing/CsvMonumentConverter.java
renchauret: going to modify the message for clarity.  I think we should still include a message though, in case the user tries to send image reference URLs without any images.  In that case, this is the only warning they would get

wrs1492 (approved):

PR #180, opened 2021-03-28
Bulk upload fixes

renchauret:
- allow commas in URLs (lists must now be split with ", " instead of ",")
- allow JPG, PNG, jpeg, JPEG images


client/src/components/Export/ExportButtons/ExportButtons.js
```diff
@@ -17,7 +17,7 @@ export default class ExportButtons extends React.Component {
             <span>
                 <span>
                     <ExportToCsvButton className="mt-2" fields={csvExportFields}
-                                       data={buildBulkExportData(monuments, csvExportFields, false)}
+                                       data={buildBulkExportData(monuments, csvExportFields, true)}
```
renchauret, line 20: This makes export add spaces after commas to be consistent with bulk upload change

client/src/components/BulkUpdateForm/BulkUpdateForm.js
```diff
@@ -129,7 +129,7 @@ export default class BulkUpdateForm extends React.Component{
 
             for (let i = 0; i < fileUpload.images.length; i++) {
                 const image = fileUpload.images[i]
-                if (!image.name.endsWith('.png') && !image.name.endsWith('.jpg')){
+                if (!image.name.endsWith('.png') && !image.name.endsWith('.jpg') && !image.name.endsWith('.PNG') && !image.name.endsWith('.JPG') && !image.name.endsWith('.jpeg') && !image.name.endsWith('.JPEG')){
```
dwang305, line 132: Could you use something like toLowerCase() for this check rather than typing out all extensions in both cases?

renchauret: no, because we don't want to allow stuff like `pNg` or `JPeg`

dwang305 (approved):

PR #184, opened 2021-04-08
Rounding longitude latitude

sxm7571:


renchauret:
Still getting "coordinates invalid" when trying to bulk upload a monument with more than 6 digits after the decimal.  I think the rounding mostly needs to be there, in the CsvMonumentConverter, so that we don't get this error anymore

renchauret (commented):
I now get this when trying to bulk upload.  Zip I'm using works on preprod and local master

renchauret (approved):

PR #185, opened 2021-04-13
Allow one folder

renchauret:
if user zips a folder with csv and images, that's fine

### DIFF
--- a/client/src/components/BulkCreateForm/BulkCreateForm.js
+++ b/client/src/components/BulkCreateForm/BulkCreateForm.js
@@ -108,7 +108,6 @@ export default class BulkCreateForm extends React.Component {
         else if (validator.matches(event.target.value, zipFileRegex)) {
             fileUpload.zip = event.target.files[0];
             let content = await JSZip.loadAsync(fileUpload.zip);
-
             for (let fileName in content.files) {
                 if (!content.files.hasOwnProperty(fileName) || fileName.startsWith('__MACOSX')) continue;
                 if (fileName.endsWith('.csv')) {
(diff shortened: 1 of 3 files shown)


client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -136,7 +135,8 @@ export default class BulkCreateForm extends React.Component {
                 const image = fileUpload.images[i]
                 if (!image.name.endsWith('.png') && !image.name.endsWith('.jpg')) {
                     if (image.name.endsWith('/')) {
-                        fileUpload.errorMessages.push('Files must be zipped directly. Do not include folders in your zip file.');
+                        continue
+                        // fileUpload.errorMessages.push('Files must be zipped directly. Do not include folders in your zip file.');
```
wrs1492, line 139: left commented code?

src/main/java/com/monumental/util/csvparsing/ZipFileHelper.java
```diff
@@ -67,7 +67,16 @@ public static File convertZipEntryToFile(ZipFile zipFile, ZipEntry zipEntry) thr
...
-        String pathName = tempDirectoryPath + "/" + zipEntry.getName();
+        String[] nameArray = zipEntry.getName().split("/");
+        String name = nameArray[nameArray.length - 1];
+
+        String pathName = tempDirectoryPath;
+        if (zipEntry.getName().startsWith("/") || zipEntry.getName().startsWith("\\")
+                || tempDirectoryPath.endsWith("/") || tempDirectoryPath.endsWith("\\")) {
+            pathName += name;
+        } else {
+        pathName += "/" + name;
```
wrs1492, line 78: indentation

wrs1492 (commented):
Besides other comments, there where some other small problems,
1. The folder is counted as an image in the uploaded photos section. I'm not sure how the backend handles extra images, but it could cause a problem if it doesn't throw them out. It could be removed from the listing, but not a big deal if it isn't
2.  the error checking isn't great for the folders. No detection on if there are multiple folders, or if there are nested folders.
3.  How is the image name read/saved? I think the path is just appended right now, which is probably not ideal to save.

renchauret:
After further testing, @wrs1492's other comments all work fine as is

wrs1492 (approved):

PR #187, opened 2021-04-27
Search By State

wrs1492:


src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -290,7 +290,9 @@ private void buildSearchQuery(CriteriaBuilder builder, CriteriaQuery query, Root
             predicates.add(this.buildSimilarityQuery(builder, query, root, searchQuery, threshold, sortByRelevance));
         }
 
-        if (latitude != null && longitude != null && distance != null) {
+        if(state != null && distance < 0) {
```
renchauret, line 293: why should distance be negative?

wrs1492: when searching by state, distance is set to negative 1, and state is now always passed with the location. Kind of a hack, but i think it is fine as it is, can rewrite if necessary

renchauret (approved):

src/main/java/com/monumental/services/MonumentService.java
renchauret: fine by me, approved, but maybe just add a quick comment to this line explaining that?

PR #188, opened 2021-05-02
add link to bulk upload instructions

renchauret:


wrs1492 (approved):
Code is good, the instructions could use a bit more detail, but is good enough. I'm also gonna throw in a change to the instructions to add the spaces between the fields since i noticed it here

PR #190, opened 2021-10-13
Clarity of Errors & Warnings

hxl1116:
Added additional table headers to provide more descriptive result data and tweaked some SCSS to account for the new columns.

### DIFF
--- a/client/src/components/BulkCreateForm/BulkCreateForm.js
+++ b/client/src/components/BulkCreateForm/BulkCreateForm.js
@@ -515,6 +515,12 @@ export default class BulkCreateForm extends React.Component {
 
         results = results.filter(result => result.errors.length > 0 || result.warnings.length > 0);
 
+        /**
+         * Get the tooltip message based on the error/warning message
+         *
+         * @param msg error/warning message
+         * @returns {string} the tooltip message
+         */
         const getTooltip = (msg) => {
             // console.log(msg)
             switch (msg) {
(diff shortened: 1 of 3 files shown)


client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -516,61 +553,83 @@ export default class BulkCreateForm extends React.Component {
...
+                        {results.map(result => {
+                            let suggestion = result['monumentSuggestion']
+
+                            return (
+                                <tr key={result.index}>
+                                    <td>{result.index}</td>
+                                    <td>{suggestion.title}</td>
+                                    <td>{suggestion.artist}</td>
+                                    <td>{suggestion.address}</td>
+                                    <td>{suggestion.date.split('T')[0]}</td>
```
stc3245, line 576: This will crash if one of the monuments a user is uploading doesn't have a value in the date field. Just need to add a check to see if it's null. 
image

client/src/components/BulkCreateForm/BulkCreateForm.js
```diff
@@ -507,6 +515,35 @@ export default class BulkCreateForm extends React.Component {
...
+            }
+        }
+
+        /**
+         * Helper function for composing the tooltip
+         *
+         * @param props OverlayTrigger properties
+         * @returns {JSX.Element} the Tooltip
+         */
+        const renderTooltip = (props) => (
```
stc3245, line 541: When compiling the front-end, there are a few warnings about variables that are defined but never used. The easiest way to rectify this would just be commenting any tool tip code out and putting a TODO in somewhere saying to fix it at some point. Alternatively, we could make a separate branch for it. 
image

client/src/components/BulkCreateForm/BulkCreateForm.js
hxl1116: Good catch, I gave it a default for when there's no value

PR #226, opened 2022-10-19
Dequeue Monument on Save

laodev:
## Description

* Remove monument from queue when bulk edit save is successful
* Update the results list to include the updated monument and render it to the result list
* Show toasts when the result of the save is received

## Link to Jira Ticket

https://colam-m.atlassian.net/browse/CMM-47
https://colam-m.atlassian.net/browse/CMM-46


dn3531 (approved):
Save works perfect. Other two buttons are WIP. Did find an unrelated bug when testing tho.

PR #251, opened 2022-11-22
fix deleting and creating tags through bulk edit

laodev:
Jira Ticket: https://colam-m.atlassian.net/browse/CMM-67?atlOrigin=eyJpIjoiZmE1ZDFjMWQ3ODdhNDg3Yzg5NmU2NTM2MzhjMmE2ODAiLCJwIjoiaiJ9

### DIFF
--- a/client/src/components/AdminPanel/BulkEdit/BulkEditPanel.js
+++ b/client/src/components/AdminPanel/BulkEdit/BulkEditPanel.js
@@ -1,5 +1,6 @@
 import React, {createContext, useCallback, useEffect, useState} from 'react'
 import {Card, Col, Container, Row} from "react-bootstrap";
+import * as QueryString from 'query-string';
 
 import SearchPanel from "./Search/SearchPanel";
 import SearchPanelBtns from "./Search/SearchPanelBtns";
@@ -12,6 +13,7 @@ import './BulkEdit.scss'
 import {QueueResetContext} from "../../../utils/queue-util";
 import { post, put } from '../../../utils/api-util';
 import EditHistoryPanel from './History/EditHistoryPanel';
+import { pick } from 'query-string';
 
 const BulkEditPanel = (props) => {
     // Hook for maintaining search results state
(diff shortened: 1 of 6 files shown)


src/main/java/com/monumental/services/TagService.java
```diff
@@ -93,30 +93,11 @@ public Tag createTag(String name, List<Monument> monuments, Boolean isMaterial)
...
      * @return Tag - The updated Tag with the specified Monument removed
      */
-    public Tag removeTagFromMonument(Tag tag, Monument monument) {
+    public Tag removeTagFromMonument(MonumentTag tag, Monument monument) {
         if (tag == null || monument == null) {
             return null;
         }
-
-        this.initializeAllLazyLoadedCollections(tag);
-
```
dn3531, line 102: Have you tested if removing this effects other parts of the system. Not 100% sure what that previous logic was for.

src/main/java/com/monumental/services/MonumentService.java
```diff
@@ -1510,7 +1513,11 @@ public void updateMonumentTags(Monument monument, List<String> newTagNames, bool
             // Un-associate any Tags/Materials from the Monument that were associated previously and no longer are
             for (Tag currentTag : currentTags) {
                 if (!newTagNames.contains(currentTag.getName())) {
-                    this.tagService.removeTagFromMonument(currentTag, monument);
+                    List<MonumentTag> monumentTags = monument.getMonumentTags().stream().filter(mt -> mt.getTag() == currentTag).collect(Collectors.toList());
```
dn3531, line 1516: Should this list only ever be expecting a single tag?

client/src/components/AdminPanel/BulkEdit/BulkEditPanel.js
```diff
@@ -34,10 +36,12 @@ const BulkEditPanel = (props) => {
     }, [searchTerm])
 
     const saveMonument = async (monument) => {
-        put(`${window.location.origin}/api/monument/bulkupdate/${monument.id}`, monument)
+        const oldMonument = searchResults.find(mon => mon.id == monument.id)
+        const newTags = monument.monumentTags.map(elem => elem.tag.name);
+        monument.monumentTags = oldMonument.monumentTags
+        put(`${window.location.origin}/api/monument/bulkupdate/${monument.id}?newTagString=${encodeURIComponent(JSON.stringify(newTags))}`, monument)
```
dn3531, line 42: 👍

src/main/java/com/monumental/services/TagService.java
laodev: @dn3531 I changed this method so now it removes monument tags from monuments directly. So because it uses MonumentTag now rather than Tag, this method doesn't do anything. I also updated the functionality in MonumentService#updateMonumentTags() so that it doesn't break the existing functionality.

src/main/java/com/monumental/services/MonumentService.java
laodev: In theory yes, a monument should only ever have one monument tag per tag.

dn3531: can be change it from a list then to a single monument. not super pressed either way

dn3531 (approved):

PR #256, opened 2022-11-29
Dev cmm 4

nak5439-579:


laodev (approved):

dn3531 (approved):

client/src/components/CreateOrUpdateForm/CreateOrUpdateForm.js
```diff
@@ -656,6 +643,39 @@ export default class CreateOrUpdateForm extends React.Component {
             }
         }
 
+        /*
+        The following form validation is based off of a conversation with Dr. Decker where the following truth table was
+        defined for various form states of known vs unknown dates:
+        _______________________________________________________________________________
+        DateOfCreation     |DateOfUninstall    |UninstallReasonGiven   |FormStateLegal?|
+        ___________________|___________________|_______________________|_______________|
+        Unknown            |Unknown            |NotGiven               |FALSE          |
```
dn3531, line 652: The error message for this condition is confusing or could be more specific. Maybe it could be 'Both date of creation and date of un-install cannot be unknown'

dn3531: And maybe a similar message for the condition underneath

dn3531 (changes requested):
Found that when having a monument with an unknown creation date, it displays as 'None' in the monument suggestion

client/src/components/ReviewModal/CreateReviewModal/CreateReviewModal.js
```diff
@@ -33,6 +33,9 @@ export default class CreateReviewModal extends React.Component {
             case DateFormat.EXACT_DATE:
                 date = prettyPrintDate(form.date, form.dateSelectValue);
                 break;
+            case DateFormat.UNKNOWN:
```
dn3531, line 36: Seems like there is some weirdness with the updateReviewModal

-> Uninstalled date unknown is coming up as NULL change
-> changing from date unknown to a known date is not coming up as a change at all

nak5439-579: fixed

PR #282, opened 2023-03-21
Cmm 65 export user m/m to csv

nak5439-579:


src/main/java/com/monumental/controllers/SearchController.java
```diff
@@ -338,4 +338,19 @@ public List<Monument> getAllMonumentsByCreatedByName(@RequestParam(required = fa
         }
         return foundMonuments;
     }
+
+    //TODO: Once CMM-86 is merged in, this endpoint need to change its name/mapping to take over it's place
```
dn3531, line 342: Can remove this comment now

dn3531 (approved):

